PR #10, opened 2018-03-29
Function Entry Package

elodszopos:
Also a hardcoded path here in package json that will need to be fixed. Please look it over @edeustace and let me know! Thanks!


packages/function-entry/configure/src/feedback-mapper.js
```diff
@@ -0,0 +1,29 @@
+export const modelToFeedbackConfig = model => {
```
edeustace: we have something similar in `text-entry` - can we re-use that? could import like so: 
`import { feedbackMapper } from '@pie-lib/config-ui;`?

packages/function-entry/configure/src/model-config.jsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Typography } from 'material-ui';
+import { InputCheckbox } from '@pie-lib/config-ui';
+import { withStyles } from 'material-ui/styles';
+import Box from './box';
+
+class RawModelConfig extends React.Component {
+  static propTypes = {
+    classes: PropTypes.object,
```
edeustace: all `.isRequired`?

packages/function-entry/controller/src/index.js
```diff
@@ -0,0 +1,79 @@
...
+const log = debug('@pie-element:function-entry:controller');
+
+const process = (v, ignoreWhitespace) => {
+    let out = v ? v.trim() : '';
+    out = !ignoreWhitespace ? out : out.replace(/ /g, '');
+    return out;
+};
+
+const isResponseCorrect = (correctResponse, model, value) => {
+    const processedValue = process(value, model.ignoreSpacing);
```
edeustace: We'll need to do more than just a string compare here, we were talking about using `mathjs.simplify`?

packages/function-entry/controller/src/index.js
```diff
@@ -0,0 +1,79 @@
...
+
+    return processedValue  === correctResponse;
+};
+
+export function model(question, session, env) {
```
edeustace: Can we use 2 spaces for indents?

packages/function-entry/configure/src/feedback-mapper.js
elodszopos: At this point, the divergence between the two is quite substantial. There are different property names, different properties altogether and a slightly different structure. If perhaps a third component might need something similar further down the road we could maybe come up with a helper function that would help generate the structures that we need.

edeustace: 👍

PR #11, opened 2018-04-02
pie-elements/calculator

harryprimoko:


packages/calculator/configure/src/main.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { InputRadio, TwoChoice } from '@pie-lib/config-ui';
+
+export default class Main extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      twoChoice: '',
```
edeustace: I don't think this is the best name, `twoChoice` refers to the component type, but we are really interested in the value within that component. So I would suggest `mode` or `calculatorMode`?

PR #16, opened 2018-04-23
Feature/point intercept

elodszopos:
Initial draft, goes along with all the other PRs. Please let me know what you think. Obvious piece that's missing are tests. I'd first like to get confirmation that what I have is alright. If it is I'll go ahead with the unit tests too!


packages/point-intercept/configure/src/add-point.jsx
```diff
@@ -0,0 +1,35 @@
...
+export class AddPointButton extends React.Component {
+  static propTypes = {
+    classes: PropTypes.object.isRequired,
+    onAddClick: PropTypes.func.isRequired,
+  };
+
+  render() {
+    const { classes, onAddClick } = this.props;
+
+    return (<div>
```
edeustace: `div` needed?

packages/point-intercept/configure/src/add-point.jsx
```diff
@@ -0,0 +1,35 @@
...
+        Add Another Point
+      </Button>
+    </div>);
+  }
+}
+
+const styles = (theme) => ({
+  button: {
+    margin: 0,
+    'margin-bottom': theme.spacing.unit * 2,
```
edeustace: `marginBottom` instead of `'margin-bottom'`?

packages/point-intercept/configure/src/delete.jsx
```diff
@@ -0,0 +1,34 @@
...
+  static propTypes = {
+    classes: PropTypes.object.isRequired,
+    onDeleteClick: PropTypes.func.isRequired,
+    disabled: PropTypes.bool.isRequired
+  };
+
+  render() {
+    const { classes, disabled, onDeleteClick } = this.props;
+
+    return (<div>
```
edeustace: `div` needed?

elodszopos:
All the feedback has been addressed, except for the misaligned labels. I tried to read through VX components, guides and source code but nothing I did fixed the misalignment. Is there any chance you can also take a look at it @edeustace? I see that you've implemented the charting lib entirely, maybe you have some insight that I'm missing in order to fix this last piece.

Everything else should be good to go. Unit tests have been included. They are pretty extensive.

edeustace:
Going to merge and deploy - only comments are tests and maybe break down configure.jsx into a set of smaller components.

PR #20, opened 2018-05-08
Feature/point intercept fixes

elodszopos:
The configure component for point-intercept has been blown up into smaller chunks that are a lot more manageable. Also added unit tests to the configure side of point-intercept.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 .idea
+.DS_Store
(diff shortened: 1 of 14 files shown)


elodszopos:
@edeustace - their internal functionality is tested as elsewhere. As long as props getting passed down are correct it should all work correctly. Testing it further would become redundant. I did however include tests for making sure they properly render. Not too extensive, but it gives us a solid safety net.

PR #26, opened 2018-06-06
Feature/graph lines

elodszopos:
Graph-line component with multiple lines included.


packages/graph-lines/configure/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@pie-element/graph-lines-configure",
+  "version": "1.0.0",
+  "description": "",
+  "main": "src/index.js",
+  "author": "",
+  "dependencies": {
+    "@material-ui/core": "^1.0.0-rc.1",
```
edeustace: you can bump these now to the official version (no more rc).

packages/graph-lines/docs/demo/config.js
```diff
@@ -0,0 +1,60 @@
...
+        partial: {
+          type: 'none',
+          default: 'Nearly'
+        },
+        incorrect: {
+          type: 'none',
+          default: 'Incorrect'
+        }
+      },
+      model: {
```
edeustace: do you think we can simplify this model? we have decided that we don't need to bind ourselves to the legacy model anymore. an example would be to have just {domain: {}, range: {}} and remove the `model.config` step etc.

elodszopos: Alright, will need to re-work the entire component base then though. Make sure it all works correctly, not a small change.

elodszopos: As discussed, because of the big implications elsewhere too, we'll postpone this update to a later time.

Details can be found in the issue below.

https://github.com/pie-framework/pie-elements/issues/29

elodszopos:
Unless there is anything else that's outstanding, I think this is good to go @edeustace!

PR #32, opened 2018-07-05
feature(categorize): updates

elodszopos:
Re-work the layout of elements, fix inputs, fix grid system and various other small updates

image

@evaneus not 100% as described in the ticket, but I've attached a screenshot, let me know what you think. I tried to make the component be more in-line with the corespring one. I also played around with the spacings a little to make it generally look a lot better than it used to.

Please let me know what you think!


evaneus:
👍

elodszopos:
Done.

Issue #51, opened 2018-08-01
[Categorize] No maximum for choice per row


alex-alberto:
Categories per row and Choices per row can be set more than 4 by directly entering the number and after that we get below issue-
![pkhadoaaojciloid](https://user-images.githubusercontent.com/36770983/43536969-0270fc48-958c-11e8-996e-d44dcad0a6fa.png)

Tested on http://sbox.powermylearning.org/piedemo/Category_Example.html

@edeustace @elodszopos @narayankj

elodszopos:
Should be resolved if the above PR that was merged has been published.

edeustace:
@alex-alberto released in `@pie-element/categorize@2.3.0` please re-test.

Issue #50, opened 2018-08-01
[Categorize] Choosing "below" option for label throws error


alex-alberto:
Click on Label Position and choose Below option, We will get below issue-
![ebahdoilnhcoeiml](https://user-images.githubusercontent.com/36770983/43536869-bb4ea4f0-958b-11e8-9353-ba4d6849c023.png)

Tested on http://sbox.powermylearning.org/piedemo/Category_Example.html

@edeustace @elodszopos @narayankj

elodszopos:
Should be resolved if the PR merged above has been published.

edeustace:
@alex-alberto fixed in `@pie-element/categorize@2.3.0` please re-test

alex-alberto:
Thanks @edeustace, it works. Closing.

PR #54, opened 2018-08-02
fix(categorize): fix changing label positions

elodszopos:
Removed yarn lock files too.


elodszopos:
Closes #50

elodszopos:
Will also close #51 now.

PR #96, opened 2018-10-29
feat(ch497): updated the partial scoring for 3 packages: match, select-text and multiple-choice

lakatosandrei:
@edeustace 
Needs to be merged only after [this](https://github.com/pie-framework/pie-lib/pull/43) pr is merged.

### DIFF
--- a/packages/match/controller/src/index.js
+++ b/packages/match/controller/src/index.js
@@ -10,19 +10,36 @@ const getResponseCorrectness = (
   const allowPartialScores = model.allowPartialScoring;
   const partialScoring = model.partialScoring;
   const rows = model.rows;
-  let totalCorrectAnswers = 0;
-  let correctAnswers = 0;
 
   if (!answers || Object.keys(answers).length === 0) {
-    return {
-      correctness: 'empty',
-      score: 0
-    };
+    return 'empty';
   }
 
-  if (rows.length) {
-    totalCorrectAnswers += rows.length * (model.layout - 1);
+  const totalCorrectAnswers = getTotalCorrect(model);
+  const correctAnswers = getCorrectSelected(rows, answers);
+
+  if (totalCorrectAnswers === correctAnswers) {
+    return  'correct';
+  } else if (correctAnswers === 0) {
+    return 'incorrect';
+  } else if (allowPartialScores && partialScoring) {
+    return 'partial';
   }
+
+  return 'incorrect';
+};
+
+const getCorrectness = (question, env, answers) => {
+  if (env.mode === 'evaluate') {
+    return getResponseCorrectness(
+      question,
+      answers
+    );
+  }
+};
+
+const getCorrectSelected = (rows, answers) => {
+  let correctAnswers = 0;
 
   rows.forEach(row => {
     const answer = answers[row.id];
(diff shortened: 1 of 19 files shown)


packages/multiple-choice/controller/src/index.js
```diff
@@ -72,28 +72,31 @@ export function outcome(config, session, env) {
...
-        resolve({ score: 0 });
+    const choices = (session.value || []).reduce((obj, c) => {
+      obj[c] = true;
+
+      return obj;
+    }, {});
+    const maxScore = config.choices.length;
+    let score = maxScore;
+
+    config.choices.forEach((c) => {
```
edeustace:
```suggestion
  const chosen = c => !!(session.value || []).find( v => v === c.value);
  const correctAndNotChosen = c => isCorrect(c) && !chosen(c);
  const incorrectAndChosen = c => !isCorrect(c) && chosen(c);
  const correctCount = config.choices.reduce( (total, choice) => {
    if( correctAndNotChosen(c) || incorrectAndChosen(c) ) {
      return total - 1;
    } else {
      return total;
  }, config.choices.length );
```

packages/multiple-choice/controller/src/index.js
```diff
@@ -72,28 +72,31 @@ export function outcome(config, session, env) {
...
+        (isCorrect(c) && !choices[c.value]) ||
+        (!isCorrect(c) && choices[c.value])
+      ) {
+        score -= 1;
       }
+    });
+
+    score = (score / maxScore).toFixed(2).replace(/[.,]00$/, '');
+
+    if (!config.partialScoring && score !== '1') {
```
edeustace:
```suggestion
    if (!config.partialScoring && correctCount  <  config.choices.length ) {
      resolve({score: 0});
    } else {
     const scoreString = ( correctCount / config.choices.length ).toFixed(2);
      resolve( {score: parseFloat( scoreString ) });
    }
```

packages/match/controller/src/index.js
```diff
@@ -19,26 +20,30 @@ const getResponseCorrectness = (
...
     }
   });
 
-  if (rows.length === correctAnswers) {
+  if (totalCorrectAnswers === correctAnswers) {
     return { correctness: 'correct', score: '100%' };
   } else if (correctAnswers === 0) {
     return { correctness: 'incorrect', score: '0%' };
-  } else if (allowPartialScores && partialScores && partialScores.length) {
+  } else if (allowPartialScores && partialScoring) {
```
edeustace: scoring logic needs to be in the `outcome` method, if the ui wants it too in evaluate mode it can call `outcome`, note that outcome should only return a value between 0.0 - 1.0, so if the ui wants a string % you can convert it in the model fn.

packages/select-text/controller/src/index.js
```diff
@@ -55,28 +56,23 @@ export const outcome = (question, session, env) => {
...
-        const rule = question.partialScoring.find(
-          p => p.numberOfCorrect === count
-        );
-        if (rule) {
-          return rule.scorePercentage / 100;
-        } else {
-          return 0;
-        }
+
+        return (count / totalCorrect.length).toFixed(2).replace(/[.,]00$/, '');
```
edeustace: returning a string score in `outcome`? should be a value between 0.0 - 1.0

PR #97, opened 2018-11-02
Added possibility to customize some of the elements from the design tab (ch557)

lakatosandrei:
This pr needs to be updated and merged only after [this](https://github.com/pie-framework/pie-lib/pull/47) pr is merged and released.

### DIFF
--- a/packages/multiple-choice/configure/src/index.js
+++ b/packages/multiple-choice/configure/src/index.js
@@ -9,8 +9,28 @@ import ReactDOM from 'react-dom';
 import Root from './root';
 import debug from 'debug';
 import { choiceUtils as utils } from '@pie-lib/config-ui';
+import defaults from 'lodash/defaults';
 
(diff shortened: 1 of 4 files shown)


packages/multiple-choice/configure/src/index.js
```diff
@@ -9,17 +9,37 @@ import ReactDOM from 'react-dom';
...
+    enableSelectChoiceLabel: true,
+    enableSelectResponseType: true,
+    enableAddChoice: true,
+    enableAddFeedBack: true,
+    enableDeleteChoice: true
+  };
+
+  return {
+    ...model,
+    config: defaults(model.config, defaultValues)
```
edeustace: can you use `model.configure` - 'configure' has a special meaning in the pie domain.

lakatosandrei: Done @edeustace

Issue #78, opened 2018-09-05
multiple-choice configure - model updated event needs to set reset appropriately


edeustace:
when you change from choiceMode checkbox -> radio, the model updated event should set `reset` to `true`. so that any preview is reset. There may be other model changes that would require a `reset: true` also.

lakatosandrei:
Fixed by [this](https://github.com/pie-framework/pie-elements/pull/101)

PR #101, opened 2018-11-08
fix(multiple-choice-item-preview-reset): resetting the item preview when changing response type

lakatosandrei:
Fixing [this](https://github.com/pie-framework/pie-elements/issues/78)


lakatosandrei:
@edeustace 
Added tests

PR #111, opened 2018-12-06
fix(multiple-choice): fixed the way the multiple-choice handles model updates

lakatosandrei:


lakatosandrei:
Found a better way with @edeustace

PR #114, opened 2019-01-21
feat: typescript -> json schema

evaneus:
See the readme in `pie-models/README.md`

Added json schemas so we can generate documentation and support data validation.

These are created through typescript definitions for question type models and config model.  The types are converted to json schemas and can copied and committed with the pie package.


packages/pie-models/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+
+### Typescript
+
+Using the types will provide IDE auto-complete and type-checking on model objects:
+
+`npm install @pie-element/pie-models`
+
+```ts
+import {MultipleChoiceConfig} from `@pie-element/pie-models`;
```
edeustace: Sorry - I missed this pr .. 
wouldn't it be better if we have: 
```javascript
import { mc } from '@pie-lib/multiple-choice';

let myConfig = types.Configure;
```

Aka put the types in the relevant package. These will be imported anyway so you won't have to add an extra import just for the type defs?

The type defs should not be part of the git repo, but are generated when  we run `yarn build` or `yarn release`.

PR #126, opened 2019-03-07
chore(general): allow empty model initialization 

### LINKED ISSUE #124: All config Elements should  be initializeable  with an empty model object.
For each configure element defined in `pie-elements/packages/[pie-name]/configure`

Verify that the UI can load with an empty `model` object. If this object is not provided the UI should  render with some sensible defaults.  

Have a conversation with @edeustace to get started on this.

sergiulucaci:
This PR aims to resolve #124 ticket.

Since this is my first contact with all packages, there might be some default values that you expect to see but ignored by me, so please review carefully default values chosen.

Besides this, there are two UI panels that seem not to work properly:
* math-inline (I cannot click inside the textarea)
* text-entry (nothing shows up, see this: https://www.dropbox.com/s/3344mag0pv38j3t/Screenshot%202019-03-07%2014.31.53.png?dl=0)

and I'm not sure if those are known issues or WIP, but if not, we could create tickets for those.

For some packages I avoided to use arrow-body-style for consistency purposes.


edeustace:
thanks @sergiulucaci will look in the morning!

evaneus:
thanks @sergiulucaci can you check with @elodszopos about math-inline?

elodszopos:
Verified with @sergiulucaci that the math-inline textarea in question is working as intended. No problems with the interaction.

PR #127, opened 2019-03-11
Chore/verified model properties

andreeapescar:
I've added some of the differences or other things that I observed.

**categorize**
- choice.categoryCount and choice. correctResponseCount - _Added on model_ - I haven't added description because I'm not sure what type should they have and what do they do.

**function-entry**

**graph-lines**
- has minimumWidth property declared in generate.js, but this property is never used - _Not added in model._
- has partialScoring property that is treated as a boolean(in pie-lib), but is treated as an array (configure.jsx - onPartialScoringChange function) as well. - _Added in model:_ another property for partialScoringRules(assuming that in the future we will use that property) as an array and partialScoring will be boolean.
- has sigfigs property but I cannot find it’s usage… I’m not sure what is this about and what type should it have - _Added in model without description._
- has showCoordinates, showPointLabels, showInputs, showAxisLabels, showFeedback properties but I cannot find their usages inside the code (some of them are displayed as checkbox, like showPointLabels, but checking/unchecking does not seem to make any change to the graph). I _haven’t removed them_ from model because I think that they will be used.

**extended-text-entry**

**inline-choice**
- has prompt property on model, but this property is never used. I _haven’t removed it_ because I think that this property will be used. - Remaining in the model
- in pie-ui has few properties that are not used yet (probably because partialScoring does not work yet) -
    - config{ responses, expression, equationEditor }
    - view
    _ Not added in the model_

**match**

**math-inline**

**multiple-choice**
- has partialScoringLabel in generate.js but property does not seem to be used. _Added in the model_ assuming that it will be used.

**number-line**
- has property tickLabelOverrides but I cannot find the usage and I am not sure about its type. _Added in model_ as array of strings.

**placement-ordering**
- has choices[i].shuffle but I could not find it’s usage. _I declared it inside the model as boolean._
-  has correctResponse[i].weight but I found this note: “Note: weights are not configurable in the existing component so we'll want do disable this in the controller and ignore it for now“. _I declared it_ inside the model.
- has partialScoring, but if I set it as a boolean crashes when I select ‘evaluate’ and if I set it as an array crashes: Invalid prop `partialScoring` of type `array` supplied to `PartialScoringConfig`, expected `boolean`. _I will set partialScoring as a boolean_ for now.
- has defaultLang but it does not seem to be used. _I will not add it in the model._
- adds these properties within the controller:
    - disabled
    - correctness
    - completeLength
    - outcomes
- has configure.tileSize but it does not seem to be used for the moment. _I will not add this_ since I am not sure about it’s value type.

**point-intercept**
- has model.config.labelsType but I cannot find it’s usage. _I will add it_ in the model having union between ‘present’ and string type, since it’s declared as ‘present’. - _I need to add description for this property._
- has minimumWidth but I cannot find it’s usage. _I will not add it in the model._

**protractor**
- does not seem to need a model. **Am I wrong?**

**ruler**

**select-text**

**text-entry**


packages/pie-models/src/pie/ruler/index.ts
```diff
@@ -0,0 +1,38 @@
+import {PieModel} from '../../PieModel';
+import { PromptConfig } from '../../PromptConfig';
+import { CommonConfigSettings } from '../../CommonConfigSettings';
+
+type RulerPieMetric = 'mm' | 'cm' | 'm' | 'km';
```
edeustace: why not just `Metric` and `Imperial`?

andreeapescar: For no reason, thank you for the suggestion. I will update it.

PR #129, opened 2019-03-11
feat(general): initialize default model

sergiulucaci:
This PR aims to resolve #124 ticket.

Default config initialization lives in a static method and the default properties are the minimum required for the UI to display and function properly.

Both config and controller methods are called `createDefaultConfig()` and default properties can be overwritten by passing the `model` as a parameter.

**Note**: I replaced the previous branch called: `chore/elements-model-initialization` with this one in order to have a cleaner commit history.


sergiulucaci:
An issue occurred related to `extended-text-entry` package. The UI panel doesn’t render due to a runtime error: `regeneratorRuntime is not defined.`

This error occurred during https://github.com/pie-framework/pie-elements/pull/121/files merge.

As a workaround I installed a new dev dependency `@babel/plugin-transform-runtime: ^7.0.0` and use it into babel.config.js plugins as follows: 

```
// babel.config.js 
…
   plugins: [
     '@babel/plugin-proposal-class-properties',
     '@babel/plugin-proposal-export-default-from',
     '@babel/plugin-proposal-export-namespace-from',
     '@babel/plugin-transform-runtime' <-- add this plugin
   ],
…
```

If you think this workaround is the right way to do it, I can submit a pull request with the fix.

sergiulucaci:
Another issue that I found is one related to the `text-entry` package. 

By default, `allowIntegersOnly` value from the controller model equals to false, which means the `Input` from `~/pie-ui/packages/text-entry/src/input.jsx` cannot render. This is because its `inputComponent` prop is null - see `getFormatTag()` method; [while a string or a class/function is expected](https://material-ui.com/api/outlined-input/#props). 

In this case I suggest to use `inputComponent` only if the returned value from `getFormatTag()` is different than null.

The changes will be made into `~/pie-ui/packages/text-entry/src/text-entry.jsx` and will look like:

https://www.dropbox.com/s/l7gzt2gef0rzaqx/Screenshot%202019-03-12%2015.30.14.png?dl=0 

Is this the right approach?

edeustace:
@sergiulucaci - re extended-text-entry build - yes add this to the root level babel config - thanks

edeustace:
@sergiulucaci re inputComp error ... 
```javascript
                inputComponent={
                  inputComponent === null || inputComponent === undefined
                    ? undefined
                    : inputComponent
                }
```
would work?

sergiulucaci:
> @sergiulucaci re inputComp error ...
> 
> ```js
>                 inputComponent={
>                   inputComponent === null || inputComponent === undefined
>                     ? undefined
>                     : inputComponent
>                 }
> ```
> 
> would work?

Yeah `undefined` does the trick. 

Would you like to have the conditional under a `constant` similar to the screenshot above or would you like to have it into the `inputComponent` property?

edeustace:
i don't mind @sergiulucaci whatever you think reads best

PR #138, opened 2019-03-15
feat(general): initialize default model (2)

sergiulucaci:
Allow initializing elements without a model or with an empty one. 

There are now three possibilities to initialize a default model:

1. Construct a Configure element without passing a model 
2. Construct a Configure element by passing an empty model
3. Construct a Configure element by calling `createDefaultModel()` static method, like:
```
const el = document.querySelector("categorize-element-configure");
const model = el.constructor.createDefaultModel(); // or you can pass a custom model:  el.constructor.createDefaultModel({ // ... custom model })
el.model = model;
```

Minimum default dummy values are used for now but please review them carefully.

Dummy values are set only in config, let me know if you want them in controllers as well.

Take note that I didn't make any changes to `graph-lines` and `point-intercept` packages, since they'll be changing soon.


evaneus:
looks like (just snapshot) tests failing

evaneus:
👍

PR #140, opened 2019-03-22
feat: ebsr implementation

sergiulucaci:
Since this package uses two instances of `@pie-element/multiple-choice/configure` we have a setting tab for each instance. If you want to have just a single setting for both instances, or if you want one for common behaviour and other two for part-specific behaviour, let me know.

Take note that this PR doesn't include tests, default initialization and documentation yet.


sergiulucaci:
Functionality, default model initialization and units tests implemented. @edeustace let me know what you think about it.

PR #145, opened 2019-04-03
fix: Update state.model if there are any changes.

andreeapescar:
Fix for https://app.clubhouse.io/keydatasystems/story/1077/multiple-choice-setting-model-property-on-config-element-doesn-t-update-ui.


evaneus:
a test would be nice, if it's not an unreasonable amount of effort to add one.

PR #148, opened 2019-04-05
Fix/multiple choice

andreeapescar:


edeustace:
Merging - see for some work re removing state: https://app.clubhouse.io/keydatasystems/story/1137/remove-componentwillreceiveprops-calls-in-root-level-react-elements

PR #155, opened 2019-04-08
Fix/placement ordering update

andreeapescar:
This has to be updated in order to use last version of pie-ui/placement-ordering.

### DIFF
--- a/packages/placement-ordering/configure/src/__tests__/design.test.jsx
+++ b/packages/placement-ordering/configure/src/__tests__/design.test.jsx
@@ -1,70 +1,295 @@
 import { shallow } from 'enzyme';
 import React from 'react';
-import Design from '../design';
-import defaultValues from '../defaultConfiguration';
+import _ from 'lodash';
 
-describe('design', () => {
-  let w;
-  let onChange;
-  let onModelChange;
+import {
+  TwoChoice,
+  FeedbackConfig,
+  FormSection,
+  InputCheckbox,
+  InputContainer,
+} from '@pie-lib/config-ui';
+import { get, set } from 'nested-property';
+import FormControlLabel from '@material-ui/core/FormControlLabel';
+import { Design } from '../design';
+import defaultValues from '../defaults';
+import { ChoiceType } from '../choice-type';
 
-  const getModel = () => ({
-    tokens: [],
-    configure: defaultValues
-  });
+jest.mock('@pie-lib/config-ui', () => ({
+  TwoChoice: props => (<div/>),
+  FeedbackConfig: props => (<div/>),
+  FormSection: props => (<div/>),
+  InputCheckbox: props => (<div/>),
+  InputContainer: props => (<div/>),
+}));
+
+jest.mock('nested-property', () => ({
+  set: jest.fn((model, modelPath, v) => {
+    model[modelPath] = v;
+
+    return model;
+  }),
+  get: jest.fn((value, valuePath) => {
+    switch (valuePath) {
+      case 'value':
+        return value[valuePath];
+      case 'taget.value':
+        return value.target.value;
+      default:
+        return value;
+    }
+  }),
+}));
+
+describe('Placement Ordering', () => {
+  let updateModel;
+  let model;
+
   beforeEach(() => {
-    onChange = jest.fn();
-    onModelChange = jest.fn();
-    w = shallow(
-      <Design
-        model={getModel()}
-        classes={{}}
-        className={'foo'}
-        onChange={onChange}
-        onModelChange={onModelChange}
-      />
-    );
+    updateModel = jest.fn();
+    model = _.cloneDeep(defaultValues);
   });
 
   describe('snapshot', () => {
-    it('renders all items with defaultProps', () => {
-      expect(w).toMatchSnapshot();
-    });
-
-    it('renders all items except feedback', () => {
-      const defaultModel = getModel();
-
-      defaultModel.configure.enableFeedback = false;
-
+    it ('renders all default items', () => {
       const wrapper = shallow(
         <Design
-          model={getModel()}
+          model={model}
           classes={{}}
           className={'foo'}
-          onChange={onChange}
-          onModelChange={onModelChange}
+          updateModel={updateModel}
         />
       );
 
       expect(wrapper).toMatchSnapshot();
     });
 
-    it('renders all items except the prompt input', () => {
-      const defaultModel = getModel();
-
-      defaultModel.configure.enablePromptChange = false;
+    it ('renders custom items', () => {
+      model.configure.settingsItemStemChange = false;
+      model.configure.settingsRemoveTileAfterPlacing = true;
 
       const wrapper = shallow(
         <Design
-          model={getModel()}
+          model={model}
           classes={{}}
           className={'foo'}
-          onChange={onChange}
-          onModelChange={onModelChange}
+          updateModel={updateModel}
         />
       );
 
       expect(wrapper).toMatchSnapshot();
     });
   });
+
+  describe('logic', () => {
+    let w;
+    let modelFn;
+
+    beforeEach(() => {
+      const wrapper = (extras) => {
+        const defaults = {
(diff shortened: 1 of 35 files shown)


packages/placement-ordering/configure/src/__tests__/design.test.jsx
```diff
@@ -1,70 +1,189 @@
...
+          updateModel={updateModel}
           onModelChange={onModelChange}
         />
       );
 
       expect(wrapper).toMatchSnapshot();
     });
   });
+
+  describe('logic', () => {
```
edeustace: looking good - can you group your tests though?
I'd suggest: 
```javascript
describe('Design', () => {
  describe('logic', () => {
    // for each function have a 1:1 mapping below, with `it`s within them
    describe('applyUpdate', () => {});
    describe('changeHandler', () => {});
    describe('onPromptChange', () => {});
    // ... plus all the rest 1:1

  });
})
```

packages/placement-ordering/configure/src/main.jsx
```diff
@@ -65,16 +65,17 @@ class Main extends React.Component {
...
+            <Tab label="Design"/>
+            <Tab label="Scoring"/>
           </Tabs>
           <Help />
         </div>
         {index === 0 && (
           <Design
             model={model}
-            onModelChange={this.onModelChange}
+            onModelChanged={this.onModelChanged}
```
edeustace: `updateModel` and `onModelChange` this looks odd, 1. there should be only 1 function that updates the model, 2. try and keep naming consistent, for a callback use `onThingVerb` eg: `onPlaneFlew` `onModelChanged` etc.

packages/placement-ordering/configure/src/design.jsx
```diff
@@ -16,57 +16,35 @@ import TextField from '@material-ui/core/TextField';
...
   constructor(props) {
     super(props);
 
-    this.state = {
-      allMoveOnDrag: false
-    };
-
     this.applyUpdate = modelFn => {
-      const { model, onModelChange } = this.props;
+      const { model, updateModel } = this.props;
```
edeustace: are you linting? this is missing in propTypes

packages/placement-ordering/configure/src/main.jsx
```diff
@@ -19,44 +19,44 @@ const styles = {
...
 
-  constructor(props, context) {
-    super(props, context);
+  constructor(props) {
+    super(props);
 
     this.state = {
-      index: 0,
-      model: props.initialModel
+      model: props.model,
```
edeustace: not part of your pr - but can we remove state?

packages/placement-ordering/configure/src/main.jsx
```diff
@@ -65,16 +65,17 @@ class Main extends React.Component {
       <div>
         <div className={classes.tabBar}>
           <Tabs onChange={this.onTabIndexChange} value={index}>
-            <Tab label="Design" />
-            <Tab label="Scoring" />
+            <Tab label="Design"/>
+            <Tab label="Scoring"/>
```
edeustace: this can be removed now no?

packages/placement-ordering/configure/src/settings.jsx
```diff
@@ -0,0 +1,108 @@
+import React from 'react';
+import Switch from '@material-ui/core/Switch';
+import FormControlLabel from '@material-ui/core/FormControlLabel';
+import { ChoiceType } from './choice-type';
+
+const onChoiceLabelAreaChange = ({ model, updateModel }) => {
```
edeustace: alot of boilerplate here - can we clean up?

packages/placement-ordering/configure/src/__tests__/design.test.jsx
```diff
@@ -1,70 +1,189 @@
 import { shallow } from 'enzyme';
 import React from 'react';
-import Design from '../design';
+import lodash from 'lodash';
+
+import {
+  TwoChoice,
```
edeustace: lint warnings here - you see them?

packages/placement-ordering/configure/src/choice-editor.jsx
```diff
@@ -21,33 +20,80 @@ function findFreeChoiceSlot(choices) {
...
     classes: PropTypes.object.isRequired,
     correctResponse: PropTypes.array.isRequired,
     choices: PropTypes.array.isRequired,
     onChange: PropTypes.func.isRequired,
-    imageSupport: PropTypes.shape({
-      add: PropTypes.func.isRequired,
-      delete: PropTypes.func.isRequired
-    }),
-    configuration: PropTypes.object
+    imageSupport: PropTypes.oneOfType([
```
edeustace: imageSupport isnt a `boolean` - why do we say it is here?

packages/placement-ordering/configure/src/choice-editor.jsx
```diff
@@ -21,33 +20,80 @@ function findFreeChoiceSlot(choices) {
 
 const log = debug('@pie-element:placement-ordering:configure:choice-editor');
 
+function updateResponse(response, from, to) {
```
edeustace: this looks very similar to some pre-existing logic we have - we can't reuse?

packages/placement-ordering/configure/src/choice-editor.jsx
```diff
@@ -1,14 +1,13 @@
-import { InputSwitch } from '@pie-lib/config-ui';
-
-import AddCircle from '@material-ui/icons/AddCircle';
 import ChoiceTile from './choice-tile';
-import IconButton from '@material-ui/core/IconButton';
 import PropTypes from 'prop-types';
 import React from 'react';
-import compact from 'lodash/compact';
 import debug from 'debug';
 import { swap } from '@pie-ui/placement-ordering';
```
edeustace: lint error here

packages/placement-ordering/configure/src/defaultConfiguration.js
```diff
@@ -1,37 +1,65 @@
 export default {
```
edeustace: why isn't this just called 'defaults.js'?

packages/placement-ordering/configure/src/settings.jsx
```diff
@@ -0,0 +1,108 @@
...
+      numberedGuidesLabel,
+      enableImagesLabel,
+      removeTilesLabel,
+      orientationLabel,
+      partialScoringLabel,
+    }
+  } = model;
+
+  return [
+    settingsChoiceLabel && renderFormControlLabel({ key: 0, checked: model.configure.editableChoiceLabel, onChange: () => onChoiceLabelAreaChange(props), label: choiceLabel}),
```
edeustace: can we try and favour jsx when laying stuff out, it reads better i think?

packages/placement-ordering/configure/src/choice-editor.jsx
andreeapescar: There is a similar function is pie-ui
function updateResponse(state, from, to) {
  const { response, opts } = state;
  const update = cloneDeep(response);
  if (opts.includeTargets) {
    if (from.type === 'choice' && to.type === 'target') {
      update[to.index] = from.id;
      return update;
    } else if (from.type === 'target' && to.type === 'choice') {
      update[from.index] = undefined;
      return update;
    } else if (from.type === 'target' && to.type === 'target') {
      return swap(response, from.index, to.index);
    } else {
      log('do nothing to the response');
      return response;
    }
  } else {
    const fromIndex = state.response.findIndex(
      r => r !== undefined && r === from.id
    );
    const toIndex = state.response.findIndex(
      r => r !== undefined && r === to.id
    );
    log('fromIndex: ', fromIndex, 'toIndex:', toIndex);
    return swap(state.response, fromIndex, toIndex);
  }
}
but it works a little bit different and returns a different value.
I believe customising it will make it pretty complex.
What do you think?

edeustace: parked for now

PR #159, opened 2019-04-10
feat(math-inline): decimal support and partial scoring

elodszopos:


elodszopos:
Resolves math-inline part of [ch440]

PR #163, opened 2019-04-16
Andreea/ch1097/update current interactions with settings

andreeapescar:


packages/pie-models/src/pie/placement-ordering/index.ts
```diff
@@ -33,226 +33,169 @@ interface CorrectResponse {
...
 
+    /** The label for answer area if placement area is enabled */
+    targetLabel?: string;
+
+    /** Indicates if teacher instructions are enabled */
+    teacherInstructions: boolean;
+}
+
+
+export interface ConfigureItem {
```
edeustace: `ConfigureField` or `ConfigureProp`?

andreeapescar: I guess ConfigureProp. Thanks.

PR #168, opened 2019-04-18
feat(math-inline): Integrated settings panel.

andreeapescar:
integrated settings panel but stoped refactoring (renaming with new established conventions) after Elod told me that he is working on that component and that there will be lots of changes.


edeustace:
want to park this  and wait or merge @andreeapescar ?

PR #169, opened 2019-04-22
feat(math-inline): formulaV2 support for controller and relevant config panel adjustments

elodszopos:
BREAKING CHANGE: controller validation schema is now using formulaV2 approach, resulting interface is changed and partial scoring is no longer supported


elodszopos:
Please hold off, got some problems because of a recent change to math-inline.

elodszopos:
Should be alright now, however, depends on https://github.com/pie-framework/pie-lib/pull/100

elodszopos:
Will need follow-up for bumping dependencies on @pie-lib.

PR #184, opened 2019-04-26
feat: hotspot implementation

sergiulucaci:
This PR includes [hotspot functionality](https://app.clubhouse.io/keydatasystems/story/1194/hotspot-authoring-ui?team_id=6). I still have to add small changes (Abe found some issues), but let me know what you think about it, Ed. 

Take note that this PR doesn't include unit tests nor documentation yet.


packages/hotspot/configure/src/root.jsx
```diff
@@ -0,0 +1,112 @@
+import Main from './main';
+import React from 'react';
+import cloneDeep from 'lodash/cloneDeep';
+import PropTypes from 'prop-types';
+
+class Root extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
edeustace: do we need this state? can we remove? i've seen this an a few comps and i think they can all be removed

edeustace:
looks good but no tests

evaneus:
need to add schema docs - see `pie-model` to generate `pie-schema.json` etc in `docs/`

packages/hotspot/configure/src/root.jsx
sergiulucaci: so instead of using state should we have something like

```
constructor(props) {
  super(props);
  this.model = props.model;
  // other lines
}
```

did I got it well?

edeustace: remove state, move handlers to custom element, pass as a prop for rendering in root.jsx. with this we have 2 states one in the custom element and 1 in root, i'm saying only have the state in the custom element and trigger renders when necessary.

PR #237, opened 2019-05-20
fix(math-inline): no longer show response index, until we support individual correctness

elodszopos:


elodszopos:
Partially resolves [ch1566].

PR #249, opened 2019-06-11
feat: hotspot rendering support for polygons

sergiulucaci:
Implement rendering support for legacy content with correct answers defined as polygon. 

See more details [here](https://app.clubhouse.io/keydatasystems/story/1616/hotspot-support-rendering-for-legacy-content-with-correct-answers-defined-as-polygon).

### DIFF
--- a/packages/hotspot/configure/src/hotspot-container.jsx
+++ b/packages/hotspot/configure/src/hotspot-container.jsx
@@ -5,6 +5,7 @@ import Help from '@material-ui/icons/Help';
 
 import Drawable from './hotspot-drawable';
 import Button from './button';
+import UploadControl from './upload-control';
 
 const isImage = (file) => {
   const imageType = /image.*/;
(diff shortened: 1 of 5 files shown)


packages/hotspot/configure/src/hotspot-container.jsx
```diff
@@ -150,6 +139,11 @@ class Container extends Component {
              } : {}}
         >
           <div className={classes.toolbar}>
+            {imageUrl ? (
+              <div className={classes.replaceSection}>
```
edeustace: As per our style guide - can you favor using jsx instead of calling a function?
```jsx
{ imageUrl && <UploadControl label="Replace Image" className={classes.replaceButton} /> }
//...
```

sergiulucaci: Done

PR #251, opened 2019-06-12
fix(math-inline): validation logic fix for allowSpaces

elodszopos:
Solves [ch1550].


edeustace:
test?

elodszopos:
Added a test too for this scenario specifically!

PR #254, opened 2019-06-19
fix(math-inline): interaction fixes, controller logic adjustment and …

elodszopos:
…small styling changes


elodszopos:
Handles quite a couple clubhouse tickets.

PR #257, opened 2019-06-26
Andreea/ch1313/graphing charting administration

andreeapescar:
This PR is ready for review.
The only thing that still has to be updated is the version of pie-ui/graphing version


packages/graphing/configure/src/graphing-config.jsx
```diff
@@ -0,0 +1,231 @@
...
+  static propTypes = {
+    classes: PropTypes.object.isRequired,
+    model: PropTypes.object.isRequired,
+    onChange: PropTypes.func.isRequired,
+    authoringEnabled: PropTypes.bool
+  };
+
+  constructor(props) {
+    super(props);
+    const toolsArr = [
```
edeustace: you have this elsewhere too - so this is duplication that we can remove?

packages/graphing/controller/src/index.js
```diff
@@ -0,0 +1,338 @@
+import debug from 'debug';
+
+import cloneDeep from 'lodash/cloneDeep';
+import lodash from 'lodash';
+import isEqual from 'lodash/isEqual';
+import { tools } from '@pie-lib/graphing';
```
edeustace: graphing is a ui package - you're using it in a controller. this is going to cause problems

if you need access to the logic in `utils` it should be in its own package with *no* UI dependencies.

packages/graphing/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "access": "public"
+  },
+  "repository": "pie-framework/pie-elements",
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "dependencies": {
+    "@pie-ui/graphing": "file:../../pie-ui/packages/graphing"
```
edeustace: use semver

packages/graphing/controller/src/index.js
```diff
@@ -0,0 +1,338 @@
...
+export const equalPolygon = (pointsA, pointsB) => {
+  const sessAnswerPoints = lodash.uniqWith(pointsA, isEqual);
+  const withoutDuplicates = lodash.uniqWith(pointsB, isEqual);
+  const sB = lodash.orderBy(sessAnswerPoints, ['x','y'], ['asc', 'asc']);
+  const sD = lodash.orderBy(withoutDuplicates, ['x','y'], ['asc', 'asc']);
+
+  return isEqual(sD, sB);
+};
+
+export const equalCircle = (A, B, C, D) => {
```
edeustace: why not use `{root, edge}` or `{center, edge}` so `isEqual(a,b){ ... } `?

packages/graphing/controller/src/index.js
```diff
@@ -0,0 +1,338 @@
...
+  const equalRootAndEdge = isEqual(D, B) && isEqual(C, A);
+  const rAB = Math.sqrt(((B.x - A.x) ** 2) + ((B.y - A.y) ** 2));
+  const rCD = Math.sqrt(((D.x - C.x) ** 2) + ((D.y - C.y) ** 2));
+  const equalRAndRoot = isEqual(C, A) && isEqual(rAB, rCD);
+
+  return equalRootAndEdge || equalRAndRoot;
+};
+
+export const equalSine = (A, B, C, D) => {
```
edeustace: there is probably a mathematical way to compare the 2 sine waves. .. 1. you'll need to derive the equation from 2 points, then compare the equation for each

packages/graphing/controller/src/index.js
```diff
@@ -0,0 +1,338 @@
...
+
+  const studentAnswerBpY = getPoints(A, B);
+  const correctAnswerBpY = getPoints(C, D);
+  const nDif = lodash.differenceWith(studentAnswerBpY.map(s => -s).reverse(), correctAnswerBpY, isEqual);
+  const dif = lodash.differenceWith(studentAnswerBpY, correctAnswerBpY, isEqual);
+
+  return dif.length === 0 || nDif.length === 0;
+};
+
+export const equalParabola = (A, B, C, D) => {
```
edeustace: same as above should a be a mathematical way of doing this rather than building data points.

packages/graphing/configure/src/correct-response.jsx
```diff
@@ -0,0 +1,192 @@
...
+
+    allTools.forEach(t => (t.toolbar = true));
+    this.state = {
+      allTools,
+      toolbarTools: allTools.reduce((a, b) => [...a, b.type], [])
+    };
+  }
+
+  componentDidMount() {
+    this.changeToolbarTools(this.state.toolbarTools);
```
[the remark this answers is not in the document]
andreeapescar: There's not. Currently, if toolbarTools are not set, in pie-ui, no element will be displayed in toolbar.

packages/graphing/controller/src/index.js
andreeapescar: I'm not sure I understand. Anyway, to check the equality of a circle, it's not enough only to check if rootA equal with rootB and egdeA equals edgeB, because root: (0,0), edge: (1, 1) is same circle defined by root: (0, 0), edge: (-1, -1), the radius of the circle is important.

PR #259, opened 2019-06-28
Implement ICA (image-cloze-association)

sergiulucaci:
Let me know if you're OK having the config model underscored; Deryl embraced it this way because it eases the export.

If you notice anything that can be improved please shoot.


evaneus:
using snake case in the model is inconsistent with everything else, but at this point I don't have big enough of a problem with it to say no.

evaneus:
@sergiulucaci think you're missing the schema docs for this. Not as critical for this particular interaction as i think the one use case for import i already handled, so can merge without but should add a post mvp story to add them.

sergiulucaci:
> @sergiulucaci think you're missing the schema docs for this. Not as critical for this particular interaction as i think the one use case for import i already handled, so can merge without but should add a post mvp story to add them.

I was thinking to not add it till I have your feedback for "snake_case".

evaneus:
Ok, we can keep it snake_case for now.

packages/image-cloze-association/controller/src/index.js
```diff
@@ -0,0 +1,96 @@
...
+        isCorrect = false;
+      }
+    });
+  } else {
+    isCorrect = false;
+  }
+  return isCorrect;
+};
+
+const isResponsePartiallyCorrect = (question, session) => {
```
edeustace: naming a function `is****` implies you are going to return a `boolean` but you return a float. so maybe rename to what that float means? like `getPartialScore`?

packages/image-cloze-association/controller/src/index.js
```diff
@@ -0,0 +1,96 @@
...
+  const denominator = maxResponsePerZone > 1 ? totalValidResponses : responseContainers.length;
+  const str = (correctAnswers / denominator).toFixed(2);
+
+  return parseFloat(str);
+};
+
+const getScore = (config, session, env) => {
+  const { partialScoring } = config;
+
+  if (!partialScoring) {
```
edeustace:
```
const score = getPartialScore(config, session); 
return partialScoring ? score : score === 1 ? 1 : 0;
```

packages/image-cloze-association/controller/src/index.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+  return isResponsePartiallyCorrect(config, session);
+};
+
+export function outcome(config, session, env) {
+  return new Promise(resolve => {
+    log('outcome...');
+    const configCamelized = camelizeKeys(config);
+
+    if (session.answers) {
```
edeustace: this will hang if there are no session answers.. maybe change to `(session.answers || [])` or the like?

sergiulucaci:
> I'd be against the underscore names personally but we can fix later

Then let's make it post-mvp: https://app.clubhouse.io/keydatasystems/story/2364/ica-change-snake-case-to-camelcase-post-mvp

sergiulucaci:
> @sergiulucaci think you're missing the schema docs for this. Not as critical for this particular interaction as i think the one use case for import i already handled, so can merge without but should add a post mvp story to add them.

Ok, making it post-mvp might work? https://app.clubhouse.io/keydatasystems/story/2371/ica-add-schema-docs-post-mvp

evaneus:
sure.

PR #277, opened 2019-07-15
Add rationale support for several Item Types

sergiulucaci:
Add rationale support for ECR, Drag in the Blank, Inline Dropdown.

Noticed that all above interactions already had `rationale` inside docs.

Issue reported [here](https://app.clubhouse.io/keydatasystems/story/2626/several-item-types-rationale-missing-from-settings-panel-and-when-toggled-on-item-authoring).

### DIFF
--- a/packages/drag-in-the-blank/configure/src/defaults.js
+++ b/packages/drag-in-the-blank/configure/src/defaults.js
@@ -64,6 +64,11 @@ export default {
       settings: true,
       label: 'Partial Scoring'
     },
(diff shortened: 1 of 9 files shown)


edeustace:
merging - but tests would be nice

sergiulucaci:
> merging - but tests would be nice

Yes, I noticed tests are missing for all these interactions, maybe we can create a task for this purpose which IMO can be post-mvp. 

cc: @evaneus

PR #288, opened 2019-07-31
Feat/prompt updates

elodszopos:


elodszopos:
Snuck in a yarn version update too because of a security risk that git warned me about, hope that's alright @edeustace.

PR #291, opened 2019-08-08
Add “correct answer indication” controller method 

sergiulucaci:
If the result from `createCorrectResponseSession` is passed to the player via `session` setter,  question should be filled with the correct answers of the question. 

Example:

```
const player = document.querySelector('hotspot-element');

createCorrectResponseSession(question, env).then(session => {
  player.session = session;  
});
```

Requirements presented [here](https://app.clubhouse.io/keydatasystems/story/2885/add-createcorrectresponsesession-to-controller-for-all-pie-elements).

Let me know if this approach looks good to you, so I can add tests (I assume this method should require tests, right?) and clean the commit history.

### DIFF
--- a/packages/categorize/controller/src/index.js
+++ b/packages/categorize/controller/src/index.js
@@ -156,3 +156,16 @@ export const outcome = (question, session, env) => {
     });
   }
 };
+
+export const createCorrectResponseSession = (question, env) => {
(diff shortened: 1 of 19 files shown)


edeustace:
pr seems fine - i just want to check how this is supposed to be used.

sergiulucaci:
@edeustace let me know if you want me to add tests

PR #304, opened 2019-08-19
chore: remove dummy prompts for various interactions

elodszopos:
As discussed with George, these no longer make sense. This kind of documentation / assistance will be handled elsewhere.


elodszopos:
Solves [ch2769].

PR #315, opened 2019-08-30
fix(hotspot): hotspotList prop type is array, not string

elodszopos:
Small fix here. React freaked out when the "incorrect" prop type was provided.


elodszopos:
Solves [ch3571]

PR #320, opened 2019-09-09
Andreea/ch3566/from ibx item authoring not able to add choices

andreeapescar:
I was still blocked on testing the changes on pie-components-integration. However, I am 99.9% sure that my changes will fix the issue.
Updated with:
Adding a choice:
- by default its value is an empty string;
- Student will not see that new choice because it is empty;
- If author tries to drag that empty choice, in order to associate it with a response area, the drag won’t be possible and an alert box will display an info message (“You need to define a value for an answer choice before it can be associated with a response area”).
- Clicking the Add Choice button to result in focus switching to the choice you have just created, so that it is in effect prompting you to populate it with a value.
- Support workflow where you first create placeholders for a bunch of answer choices all at once, and then populate them sequentially. 

Editing a choice (Author starts editing an existing choice and removes all the characters.):
- Editing an answer choice that is not associated with any response areas, then arguably saving it as blank should be equivalent to deleting it.
- assuming that the choice was previously associated with a response area (or multiple response area if duplicates are allowed), saving it as blank shouldn’t be allowed (Display an error message, “Answer choices cannot be blank” and revert the field to the last saved value.)

### DIFF
--- a/packages/drag-in-the-blank/controller/src/utils.js
+++ b/packages/drag-in-the-blank/controller/src/utils.js
@@ -25,3 +25,5 @@ export const getAllCorrectResponses = ({ correctResponse, alternateResponses }) 
     numberOfPossibleResponses: undefined
   });
 };
+
+export const choiceIsEmpty = choice => choice && choice.value && (choice.value.trim() === '' || choice.value.replace(/<[^>]*>?/gm, '') === '');
--- a/packages/drag-in-the-blank/controller/src/utils.js
+++ b/packages/drag-in-the-blank/controller/src/utils.js
@@ -26,4 +26,12 @@ export const getAllCorrectResponses = ({ correctResponse, alternateResponses }) 
   });
 };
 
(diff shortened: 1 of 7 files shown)


packages/drag-in-the-blank/controller/src/utils.js
```diff
@@ -24,4 +24,14 @@ export const getAllCorrectResponses = ({ correctResponse, alternateResponses })
...
   });
-};
\ No newline at end of file
+};
+
+export const choiceIsEmpty = choice => {
+  if (choice) {
+    const { value = '' } = choice;
+
+    return value.trim() === '' || value.replace(/<[^>]*>?/gm, '') === '';
```
elodszopos: I would move regex outside so it doesn't get created every single time the function runs. Regex instantiation costs quite a bit of memory.

elodszopos: There's another one in `markupUtils.js`, same applies to that one as well.

PR #349, opened 2019-10-01
fix(math-inline): simple mode literal evaluation fix for decimals and spaces

elodszopos:
Solves [ch3909]


elodszopos:
Also solves [ch3912]

PR #351, opened 2019-10-04
fix(hotspot): controller scoring function fix

elodszopos:


elodszopos:
Solves part of [ch3536]

PR #356, opened 2019-10-08
fix(math-inline): fix controller logic safety net - ch3905

elodszopos:
Fixes [ch3905]


elodszopos:
Unblocks [ch3912], fixes [ch4662], unblocks [ch3679]

PR #361, opened 2019-10-14
Edeustace/ch4802/verify session loading in player

andreeapescar:
Added session files with examples for how the session should look like.
+ fixed an issue on charting caused by reseting the model.


edeustace:
see #362

PR #364, opened 2019-10-16
fix:  Ch2766 unresolved promises in controllers

evaneus:


evaneus:
@andreeapescar applied these changes in different branch.

PR #368, opened 2019-10-24
feat(controller): made sure controller is not divulging sensitive data when not in evaluate mode [ch5020]

lakatosandrei:


edeustace:
tests?

packages/match/controller/src/__tests__/index.test.jsx
```diff
@@ -110,17 +123,29 @@ describe('model', () => {
       });
     });
 
+    it('returns a new object, without using the same reference', () => {
```
edeustace: object references aren't really an issue, when run in a secure environment, the `model` function will run remotely and be returns as json - so they'll never be the same reference.

packages/match/controller/src/__tests__/index.test.jsx
```diff
@@ -131,6 +156,10 @@ describe('model', () => {
       result = await model(question, session, env);
     });
 
+    it('returns a new object, without using the same reference', () => {
```
edeustace: see above

packages/match/controller/src/__tests__/index.test.jsx
lakatosandrei: If I didn't `clone` the object the model would never come back with rows that have values. So I had to clone it, look below at my comment.

packages/match/controller/src/index.js
```diff
@@ -168,6 +170,10 @@ export function model(question, session, env, updateSession) {
 
     normalizedQuestion.rows.forEach(row => {
       correctResponse[row.id] = row.values;
+
```
lakatosandrei: @edeustace 
Because of the below statement I had to clone it.

PR #396, opened 2019-12-11
fix: ch6025: 

andreeapescar:
fix for Dichotomous scoring type: EBSR should be scored as 1/1 if both parts are correct, and as 0/1 if either part is wrong.

BREAKING CHANGE: replaced `model.partA.partialScoring` and `model.partB.partialScoring` with `model.partialScoring`, replaced `model.partA.scoringType` and `model.partB.scoringType` with `model.scoringType`, replaced `configuration.partA.partialScoring` and `configuration.partB.partialScoring` with `configuration.partialScoring`, replaced `configuration.partA.scoringType` and `configuration.partB.scoringType` with `configuration.scoringType`
feat: ch4719: Scoring settings for EBSR are not correct (partial scoring).

Updated docs.


edeustace:
Not going to release this until we figure out model change implications

PR #408, opened 2020-01-06
feat(math-inline): remove stringCheck flag from authoring

elodszopos:


edeustace:
can you remove from the controller too @elodszopos ?

elodszopos:
Updated with fix for [ch6453]. As discussed, I left the safety fallback string compare in case of parse failure. Feel free to use it to improve literal validation when you're back on working on the tidy up that you're doing.

PR #433, opened 2020-03-25
fix(math-inline): allowThousandsSeparator behavior change - ch7109

elodszopos:


elodszopos:
There seem to be some partial scoring failed tests that aren't related to my work

elodszopos:
Solves [ch7109]

PR #452, opened 2020-04-21
fix(math-inline): fix evaluation logic - stop on correct answer found - PD-66

elodszopos:


elodszopos:
Good to go @edeustace

PR #482, opened 2020-11-20
Feature/matrix

relumesaros:


evaneus (commented):
Quick comment - think you may need to add types/schema docs using pie-models?

relumesaros:
> Quick comment - think you may need to add types/schema docs using pie-models?

Hi @evaneus 
I've added the types/schema docs using pie-models

edeustace:
looks good - can you add some tests?

relumesaros:
> looks good - can you add some tests?

Added

PR #502, opened 2021-02-02
fix(multiple-choice): set color to primary

PatriciaRomaniuc:


elodszopos (approved):

elodszopos (approved):
The change needs to be universal here and should not be coming from a configuration - we need the same styles applied to all the inputs that we have of the same kind. We should use the `color` lib from `@pie-lib/render-ui`

PR #533, opened 2021-03-10
fix(passage): preserve non-breaking spaces in content in order to make indentation visible PD-611

CarlaCostea:
https://illuminate.atlassian.net/browse/PD-611

### DIFF
--- a/packages/passage/src/stimulus-tabs.jsx
+++ b/packages/passage/src/stimulus-tabs.jsx
@@ -67,8 +67,6 @@ class StimulusTabs extends React.Component {
 
   parsedText = (text) => text.replace(/(<br\/>\n)/g, '<br/>');
 
-  createMarkup = (tab) => { return { __html: this.parsedText(tab) } };
-
   render() {
     const { classes, tabs } = this.props;
     const { activeTab } = this.state;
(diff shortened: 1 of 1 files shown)


packages/passage/src/stimulus-tabs.jsx
```diff
@@ -84,7 +90,7 @@ class StimulusTabs extends React.Component {
                 label={
                   <Purpose purpose="passage-title">
                     <span
-                      dangerouslySetInnerHTML={{ __html: tab.title }}
+                      dangerouslySetInnerHTML={this.createMarkup(tab.title)}
```
andreeapescar, line 93: I would keep it like: { __html: this.parsedText(tab) } here (instead of using a function that returns an object with __html inside )

andreeapescar (changes requested):

packages/passage/src/stimulus-tabs.jsx
CarlaCostea: done

CarlaCostea: done

andreeapescar (approved):

PR #539, opened 2021-03-12
fix(ebsr): set only one correct answer when controls are switched from checkbox to radio button PD-262

CarlaCostea:
https://illuminate.atlassian.net/browse/PD-262

### DIFF
--- a/packages/ebsr/configure/src/main.jsx
+++ b/packages/ebsr/configure/src/main.jsx
@@ -29,33 +29,32 @@ export class Main extends React.Component {
   onModelChanged = (model, key) => {
     const { onModelChanged } = this.props;
 
-    switch (key) {
-      case 'partB.choiceMode': {
-        let value = model.partB.choiceMode;
+    if (key !== 'partB.choiceMode') {
+      return onModelChanged(model);
+    };
 
-        if (value === 'radio') {
-          let correctFound = false;
+    const value = model.partB.choiceMode;
 
-          model.partB.choices = model.partB.choices.map(c => {
-            if (correctFound) {
-              c.correct = false;
-              return c;
-            }
+    if (value === 'radio') {
+      let correctFound = false;
 
-            if (c.correct) {
-              correctFound = true;
-            }
-            return c;
-          });
-        }
-        onModelChanged(model, true);
-        break;
-      }
-      default:
-        onModelChanged(model);
-        break;
-    }
-  };
+      model.partB.choices = model.partB.choices.map(c => {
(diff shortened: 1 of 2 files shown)


packages/ebsr/configure/src/main.jsx
```diff
@@ -26,12 +26,42 @@ export class Main extends React.Component {
     onConfigurationChanged: PropTypes.func
   };
 
+  onModelChanged = (model, key) => {
+    const { onModelChanged } = this.props;
+
+    switch (key) {
```
andreeapescar, line 32: No need for "switch" here, you only have one different case to treat.

packages/ebsr/configure/src/main.jsx
```diff
@@ -26,12 +26,42 @@ export class Main extends React.Component {
...
+          model.partB.choices = model.partB.choices.map(c => {
+            if (correctFound) {
+              c.correct = false;
+              return c;
+            }
+
+            if (c.correct) {
+              correctFound = true;
+            }
+            return c;
```
andreeapescar, line 48: I think you can use only one "if" here:

if (c.correct && correctFound) {
c.correct = false;
correctFound = true;
}
return c;

Please correct me if this doesn't what you're expecting.

andreeapescar (changes requested):

packages/ebsr/configure/src/main.jsx
CarlaCostea: ok

packages/ebsr/configure/src/main.jsx
CarlaCostea: 'correctFound' becomes true when first correct answer is found. then all 'c.correct' are made false

packages/ebsr/configure/src/main.jsx
```diff
@@ -26,12 +26,41 @@ export class Main extends React.Component {
...
   };
 
+  onModelChanged = (model, key) => {
+    const { onModelChanged } = this.props;
+
+    if (key !== 'partB.choiceMode') {
+      return onModelChanged(model);
+    };
+
+    let value = model.partB.choiceMode;
```
andreeapescar, line 36: "value" can be const (it doesn't get updated)

andreeapescar (approved):

packages/ebsr/configure/src/main.jsx
CarlaCostea: sure

andreeapescar (approved):

PR #541, opened 2021-03-18
fix: Do not set session.value if placementArea is true.

andreeapescar:
https://user-images.githubusercontent.com/20281464/111593720-ea606580-87d2-11eb-9a66-ff3b26b31294.mov

https://user-images.githubusercontent.com/20281464/111593955-2abfe380-87d3-11eb-997c-40ed23b8fe69.mov

As you can see, previously, the session.value was always being set for placementArea: true.
Since that is incorrect, I removed it (only for placementArea: true)

### DIFF
--- a/packages/placement-ordering/controller/src/__tests__/index.test.js
+++ b/packages/placement-ordering/controller/src/__tests__/index.test.js
@@ -228,6 +228,40 @@ describe('index', () => {
         expect(noResult).toBeNull();
       });
     });
+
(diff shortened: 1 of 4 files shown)


lakatosandrei (approved):

lakatosandrei:
Please add tests

lakatosandrei (changes requested):
Please add tests

lakatosandrei (approved):

PR #546, opened 2021-03-19
[multiple-choice]: add support for horizontal layout on settings

dianagociu:
Add support for horizontal/vertical layout toggle

image


alextkd:
I fixed the snapshot and test passes fine now, I'm not sure why is not running. @andreeapescar / @elodszopos ?

alextkd:
@dianagociu I would also reset the spaces updates

packages/multiple-choice/configure/src/defaults.js
```diff
@@ -16,9 +16,14 @@ export default {
     rationaleEnabled: true,
     limitChoicesNumber: true,
     teacherInstructionsEnabled: true,
-    studentInstructionsEnabled: true
+    studentInstructionsEnabled: true,
+    verticalMode: true
   },
   configuration: {
+    verticalMode: {
+      settings: true,
```
andreeapescar, line 24: I would rather not display the toggle for Vertical Mode in the Settings Panel (should be `settings: false`).
By default, in the model, it should be true, that's correct.

andreeapescar: Also, we should add the new property in the docs. (pie-models)

packages/pie-models/src/pie/multiple-choice/index.ts
```diff
@@ -32,6 +32,11 @@ export interface MultipleChoicePie extends PieModel {
   /** Determines if prompt should show */
   promptEnabled?: boolean;
 
+  /** Indicates the layout of choices for player
+   * @default: true
+   */
+  verticalLayout?: boolean;
```
andreeapescar, line 38: I see you added **verticalMode** property, but in docs you've named it **verticalLayout**.

@dianagociu , can we please use verticalMode in the docs as well?

andreeapescar (approved):

PR #548, opened 2021-03-22
fix(math-inline): fix session value setting

elodszopos:
Since the reference was being overwritten entirely, it was freaking out - session was losing values and updates weren't being passed on.


edeustace:
looks like this was already fixed in https://github.com/pie-framework/pie-elements/commit/a5cff102a58632ae3f56c33ad0af6d25ae400a1b

PR #553, opened 2021-03-24
feat(math-inline): add allowTrailingSpaces. BREAKING CHANGE: added al…

PatriciaRomaniuc:
…lowTrailingSpaces to model PD-282
https://illuminate.atlassian.net/browse/PD-282


andreeapescar:
Let's have a new configuration property (instead of setting it directly on the model).

configuration will have a new property **allowTrailingZeros** with: settings, label, enabled & default.

If model.allowTrailingZeros is defined, we use that value. If it is not defined, we check configuration.allowTrailingZeros.default. If defined, use that value, else, use false.

Also, the checkbox has to be above all responses, instead of being in each individual response.

andreeapescar (changes requested):

andreeapescar (approved):
Looks good. Thank you

edeustace:
lets wait until math-validation is active to see if we still need this?

PR #562, opened 2021-04-19
fix: image-cloze-association create schemas for PD-1095

PatriciaRomaniuc:
https://illuminate.atlassian.net/browse/PD-1095
https://illuminate.atlassian.net/browse/PD-1114


packages/pie-models/src/pie/image-cloze-association/index.ts
```diff
@@ -0,0 +1,110 @@
...
+    rationaleEnabled?: boolean;
+
+    /** Indicates if Teacher Instructions are enabled */
+    teacherInstructionsEnabled?: boolean;
+
+    /** Indicates if Student Instructions are enabled */
+    studentInstructionsEnabled?: boolean;
+
+    /** The image over which the responses will be dragged*/
+    image?: {
```
andreeapescar, line 61: I guess we could create a new Interface for this object?

packages/image-cloze-association/controller/src/__tests__/index.test.js
```diff
@@ -323,8 +327,8 @@ describe('controller', () => {
           validResponse: {
             score: 1,
             value: [
-              [rhomb, square],
-              [rhomb, square, trapeze],
+              {images: [rhomb, square]},
```
andreeapescar, line 330: Let's add tests where we have {}, or { images: null} etc. (This will require to add sanity checks on the next file).

andreeapescar (changes requested):
Also, please make sure you commit with **BREAKING CHANGE:**  .

andreeapescar (approved):

andreeapescar:
@PatriciaRomaniuc I'm afraid we

PR #565, opened 2021-04-21
feat: image-cloze-association BREAKING CHANGE: add showDashedBorder to model PD-1114

PatriciaRomaniuc:
I had to create the branch from fix/PD-1095/create-image-cloze-association-schemas in order to update the model schema with the new property. Also, I renamed and fixed the tests because they were not executed.
https://illuminate.atlassian.net/browse/PD-1114


andreeapescar:
Would it be helpful to create a PR in this case into your branch, and merge everything into a single branch (one PR), @PatriciaRomaniuc ?

PR #586, opened 2021-05-13
fix(number-line): partial scoring not working correctly PD-448

PatriciaRomaniuc:
https://illuminate.atlassian.net/browse/PD-448

### DIFF
--- a/packages/number-line/controller/src/index.js
+++ b/packages/number-line/controller/src/index.js
@@ -47,12 +47,17 @@ export function outcome(model, session, env) {
         0
       );
 
+      let total = model.correctResponse.length;
       let numIncorrect = 0;
-      if((session.answer || []).length > model.correctResponse.length) {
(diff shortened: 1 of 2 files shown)


packages/number-line/controller/src/index.js
```diff
@@ -46,8 +46,18 @@ export function outcome(model, session, env) {
         accumulateAnswer(model.correctResponse),
         0
       );
+
+      let numIncorrect = 0;
+      if((session.answer || []).length > model.correctResponse.length) {
+         numIncorrect = (session.answer || []).length - model.correctResponse.length;
+      }
+
       const total = model.correctResponse.length;
```
andreeapescar, line 55: move total above (line 49) and use that constant instead of model.correctResponse.length

andreeapescar: also, could total be 0?

andreeapescar (approved):

PR #590, opened 2021-05-17
feat: Displayed a message when the item has one or more alternate answers - PD-167

andreeimiron:
https://illuminate.atlassian.net/browse/PD-167
Applies to:
- Explicit Constructed Response
- Categorize
- Placement Ordering


packages/explicit-constructed-response/src/main.js
```diff
@@ -71,23 +74,27 @@ export class Main extends React.Component {
...
-            </div>
-          )
-        }
         <ConstructedResponse
           {...this.props}
           onChange={this.onChange}
           showCorrectAnswer={showCorrectAnswer}
           value={value}
         />
+        {(showCorrectAnswer || env && env.mode === 'view' && env.role === 'instructor') && showNote && note && (
```
andreeapescar, line 83: use destructurong and sanity check

andreeimiron: done

andreeapescar (approved):

PR #594, opened 2021-05-25
fix(categorize): Fixed Row labels appearance and the display of categories - PD-197

andreeimiron:
https://illuminate.atlassian.net/browse/PD-197

### DIFF
--- a/packages/categorize/src/categorize/index.jsx
+++ b/packages/categorize/src/categorize/index.jsx
@@ -137,7 +137,7 @@ export class Categorize extends React.Component {
     log('[render] disabled: ', model.disabled);
 
     const { rowLabels, categoriesPerRow } = model;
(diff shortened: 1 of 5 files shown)


packages/categorize/src/categorize/index.jsx
```diff
@@ -136,7 +136,8 @@ export class Categorize extends React.Component {
 
     log('[render] disabled: ', model.disabled);
 
-    const { rowLabels } = model;
+    const { rowLabels, categoriesPerRow } = model;
+    const nbOfRows = categories && Math.ceil(categories.length / categoriesPerRow);
```
andreeapescar, line 140: Sanity checks (if categories are null or undefined, how it this going to behave?)

andreeapescar (approved):

PR #608, opened 2021-06-21
Feat(Multiple choice): Grid Layout

dianagociu:
Added new setting: choicesLayout - a dropdown with 3 possibilities : vertical, grid and horizontal - with vertical default. When we select grid for the layout a new setting will show up - grid Columns - which is also a dropdown with numbers between 1 and the number of choices . The selected number will represent the number of columns that we want for the grid. 
@alextkd


dianagociu:
image

dianagociu:
Horizontal Layout: 
image

dianagociu:
Grid Layout with 2 columns:
image

packages/multiple-choice/src/multiple-choice.jsx
```diff
@@ -204,7 +214,7 @@ export class MultipleChoice extends React.Component {
     } = this.props;
     const { showCorrect } = this.state;
     const isEvaluateMode = mode === 'evaluate';
-
+    const layoutClass = this.props.choicesLayout === 'grid' ? classes.gridLayout : (this.props.choicesLayout === 'vertical'? '' : classes.horizontalLayout)
```
alextkd, line 217: I would write this more simple and more clear using `classNames({ [classes.gridLayout]: this.props.choicesLayout === 'grid', {[classes.horizontalLayout]: this.props.choicesLayout === 'horizontal' } })`

packages/multiple-choice/src/multiple-choice.jsx
```diff
@@ -225,10 +235,11 @@ export class MultipleChoice extends React.Component {
         />
         <br />
         <PreviewPrompt className="prompt" prompt={prompt}/>
-        <div className={this.props.verticalMode ? '' : classes.horizontalLayout}>
+        <div className={classNames(layoutClass)} style={styles.getColumns(this.props.gridColumns)}>
```
alextkd, line 238: then you don't need to use it here

alextkd (changes requested):
Let's make the upper improvements

edeustace:
@andreeapescar - mind managing the merge of this - looks good to me - thanks @dianagociu and @alextkd

packages/multiple-choice/configure/src/defaults.js
```diff
@@ -17,12 +17,16 @@ export default {
     limitChoicesNumber: true,
     teacherInstructionsEnabled: true,
     studentInstructionsEnabled: true,
-    verticalMode: true
+    choicesLayout: 'vertical',
+    gridColumns: 2
   },
   configuration: {
-    verticalMode: {
+    choicesLayout: {
```
andreeapescar, line 24: Since you're removing an existing property, I think it would be better to have a breaking change.

packages/multiple-choice/configure/src/main.jsx
```diff
@@ -107,6 +108,9 @@ const Design = withStyles(styles)(props => {
     promptEnabled
   } = model || {};
 
+
+  const nrOfColumnsAvailable = Array.from({length: model.choices.length}, (_, i) => i + 1);
```
andreeapescar, line 112: We should add a sanity check here.

andreeapescar (approved):

PR #610, opened 2021-06-25
Feat(multiple choice): Toolbar position

dianagociu:
Added new setting : toolbarEditorPosition which is an enum : "bottom" / "top" with bottom default. This will indicate the position for the editor toolbar.  ( In the screenshots below I set the toolbarEditorPosition to 'top' )
Screenshot 2021-06-25 at 16 07 31
Screenshot 2021-06-25 at 16 07 54
Screenshot 2021-06-25 at 16 08 05
Screenshot 2021-06-25 at 16 08 15
@alextkd


packages/multiple-choice/configure/src/main.jsx
```diff
@@ -125,6 +125,7 @@ const Design = withStyles(styles)(props => {
             onChange={onTeacherInstructionsChanged}
             imageSupport={imageSupport}
             nonEmpty={false}
+            toolbarOpts={{ position: model.toolbarEditorPosition === 'top' ? model.toolbarEditorPosition : 'bottom' }}
```
andreeapescar, line 128: I would rather save this in a variable, since you're using it 4 times.
I think you can have something like:

```
const toolbarOpts = {};

switch (model.toolbarEditorPosition) {
   case 'top':
     toolbarOpts.position = 'top';
     break;
   default:
      toolbarOpts.position = 'bottom';
       break;
}
```

before the Content's declaration.

andreeapescar:
Also, as far as I know, there should also be a PR in pie-lib for this. Let's have that one merged first. :)

andreeapescar (approved):

edeustace:
merged locally

PR #627, opened 2021-07-13
feat(math-inline): added option to set toolbar editor position

dianagociu:
Screenshot 2021-07-13 at 14 57 18
Screenshot 2021-07-13 at 14 57 26
Screenshot 2021-07-13 at 14 57 34
Screenshot 2021-07-13 at 14 58 01
@alextkd

### DIFF
--- a/packages/math-inline/docs/demo/generate.js
+++ b/packages/math-inline/docs/demo/generate.js
@@ -17,6 +17,7 @@ const initialModel = {
 
 const E262456 = {
   'equationEditor': 3,
+  'toolbarEditorPosition': "top",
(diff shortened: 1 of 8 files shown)


packages/math-inline/docs/demo/generate.js
```diff
@@ -17,6 +17,7 @@ const initialModel = {
 
 const E262456 = {
   'equationEditor': 3,
+  'toolbarEditorPosition': 'top',
```
PatriciaRomaniuc, line 20: Set toolbarEditorPosition to 'bottom' instead of 'top'.

PatriciaRomaniuc (approved):

PR #647, opened 2021-08-05
fix(image-cloze-association): Removed correct answer in evaluate mode  for empty response

andreeimiron:


packages/image-cloze-association/src/possible-response.jsx
```diff
@@ -35,7 +35,11 @@ export class PossibleResponse extends React.Component {
 
     return connectDragSource(
       <div className={`${classes.base} ${additionalClass}`} style={containerStyle}>
-        <span className={classes.span} dangerouslySetInnerHTML={{__html: data.value}}/>
+        <span
+          style={data.hidden && { visibility: 'hidden' }}
```
andreeapescar, line 39: use classnames instead, please

PR #663, opened 2021-08-10
feat(math-inline): if absolute notation is a custom key, it should be…

PatriciaRomaniuc:
…have as the regular button PD-1257
https://illuminate.atlassian.net/browse/PD-1257


PatriciaRomaniuc:
Closed because of: https://github.com/pie-framework/pie-lib/pull/338
That fix will also include this.

PR #665, opened 2021-08-12
fix(CTT-174): add safari tooltip fix 

sergiulucaci:
In classroom-tool context the tooltip is still visible when questions are changed by teacher. 

See this: https://illuminate.atlassian.net/browse/CTT-183


sergiulucaci:
Unfortunately this hack won't work.

PR #668, opened 2021-08-16
fix(math-inline): render parallel sign with bigger font-size PD-1258

PatriciaRomaniuc:
**NEED TO MERGE https://github.com/pie-framework/pie-lib/pull/343 FIRST**

### DIFF
--- a/packages/math-inline/src/main.jsx
+++ b/packages/math-inline/src/main.jsx
@@ -61,6 +61,16 @@ function prepareForStatic(model, state) {
   }
 }
 
+// increase the font of the parallel notation
+const updateSpans = () => {
+  const spans = Array.from(document.querySelectorAll('span[mathquill-command-id]'));
+  (spans || []).forEach(span => {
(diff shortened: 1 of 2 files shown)


packages/math-inline/src/main.jsx
```diff
@@ -61,6 +61,16 @@ function prepareForStatic(model, state) {
   }
 }
 
+// increase the font of the parallel notation
+const updateSpans = () => {
+  const spans = Array.from(document.querySelectorAll('span[mathquill-command-id]'));
```
andreeapescar, line 66: Can we use the function that was defined in pie-lib in this PR: https://github.com/pie-framework/pie-lib/pull/343?

andreeapescar (approved):

packages/math-inline/src/main.jsx
andreeapescar: @PatriciaRomaniuc we have pie-lib released

PR #670, opened 2021-08-18
fix/use-literal-as-default-validation-type

CarlaCostea:
- math-validation should be published & yalc files should be removed before merge
- I can't run tests at this point, because math-validation is linked in two files. (waiting for math-validation package to be published)

Screenshot 2021-08-16 at 14 45 48


andreeapescar:
Please make sure you initialize "validation" on responses if they don't have anything set.

CarlaCostea:
please consider PR https://github.com/pie-framework/pie-elements/pull/671

PR #675, opened 2021-08-20
feat/math-inline-literal-validation-options

CarlaCostea:
My task was: 
image

In order to be consistent in configuration I use `enabled` not `controls`.

Please see documentation here: 
/Users/carla/Documents/Work/pie-framework/pie-elements/packages/pie-models/src/pie/math-inline/index.ts

Resulted behaviour is the following:

Example 1. validation literal `ignoreOrder: { enabled :true} && allowTrailingZeros: {enabled: true}`
image

Example 2. validation literal `ignoreOrder: { enabled :false} && allowTrailingZeros: {enabled: true}` && onDefault model we have ignoreOrder:true and allowTrailingZeros: true 

Design result will be this: 
image
 Validation will consider ignoreOrder option true, without possibility to change it from UI

### DIFF
--- a/packages/extended-text-entry/docs/demo/generate.js
+++ b/packages/extended-text-entry/docs/demo/generate.js
@@ -2,7 +2,7 @@ exports.model = (id, element) => ({
   id,
   element,
   customKeys: [
-    "\\square"
+    '\\square'
   ],
   feedback: { type: 'default', default: 'this is default feedback' },
   prompt: 'This is the question prompt',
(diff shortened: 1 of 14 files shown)


andreeapescar:
It looks really good now 💯 Good job!

Please make sure George approves the behaviour before merging it into your branch and fix tests.

Also, I would like to have this one merged first (into feat/link-math-validation), and just after that we can merge feat/link-math-validation, because you've made some changes in this branch that would be BREAKING changes for feat/link-math-validation otherwise, and we can avoid that. :)

PR #691, opened 2021-09-02
feat(passage): implement print PD-1282

PatriciaRomaniuc:
Rendered tabs in sequence for more than one passage:
Screen Shot 2021-09-02 at 9 57 38 AM
View with only one passage:
Screen Shot 2021-09-02 at 9 58 08 AM


packages/passage/src/stimulus-tabs.jsx
```diff
@@ -72,7 +79,26 @@ class StimulusTabs extends React.Component {
     const { activeTab } = this.state;
 
     if (tabs && tabs.length > 1) {
-      return (
+      return tabs[0].disabledTabs ? (
```
andreeapescar, line 82: I would suggest to move `disabledTabs` property on the model.

andreeapescar (changes requested):

andreeapescar (approved):

PR #693, opened 2021-09-03
feat(explicit-constructed-response): implement print PD-1284

PatriciaRomaniuc:
**_Instructor view:_** 
Screen Shot 2021-09-03 at 11 00 58 AM
**_Student view_**
Screen Shot 2021-09-03 at 11 01 11 AM

### DIFF
--- a/packages/explicit-constructed-response/src/main.js
+++ b/packages/explicit-constructed-response/src/main.js
@@ -6,6 +6,7 @@ import CorrectAnswerToggle from '@pie-lib/correct-answer-toggle';
 import { ConstructedResponse } from '@pie-lib/mask-markup';
 import { color, Collapsible, hasText } from '@pie-lib/render-ui';
 import { withStyles } from '@material-ui/core/styles';
+import classNames from 'classnames';
 
 export class Main extends React.Component {
   static propTypes = {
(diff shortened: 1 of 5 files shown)


packages/explicit-constructed-response/src/main.js
```diff
@@ -53,29 +56,44 @@ export class Main extends React.Component {
...
             <div className={classes.collapsible}>
-              <Collapsible
-                labels={{ hidden: 'Show Teacher Instructions', visible: 'Hide Teacher Instructions' }}
-              >
-                <div dangerouslySetInnerHTML={{ __html: teacherInstructions }}/>
-              </Collapsible>
+              {!animationsDisabled ? <Collapsible
+                  labels={{ hidden: 'Show Teacher Instructions', visible: 'Hide Teacher Instructions' }}
+                >
+                  <div dangerouslySetInnerHTML={{ __html: teacherInstructions }}/>
```
andreeapescar, line 84: Let's move this outside (the div with html) so if anything has to change, we can change in a single place.

packages/explicit-constructed-response/src/main.js
```diff
@@ -84,17 +102,19 @@ export class Main extends React.Component {
...
           <div className={classes.collapsible}>
-            <Collapsible
-              labels={{ hidden: 'Show Rationale', visible: 'Hide Rationale' }}
-            >
-              <div dangerouslySetInnerHTML={{ __html: rationale }}/>
-            </Collapsible>
+            {!animationsDisabled ? <Collapsible
+                labels={{ hidden: 'Show Rationale', visible: 'Hide Rationale' }}
+              >
+                <div dangerouslySetInnerHTML={{ __html: rationale }}/>
```
andreeapescar, line 114: Same here

andreeapescar (approved):

PR #740, opened 2021-10-21
feat(match): change drag-and-drop behavior from swap to push PD-1336

PatriciaRomaniuc:
https://illuminate.atlassian.net/browse/PD-1336


packages/match/configure/src/answer-config-block.jsx
```diff
@@ -97,7 +96,13 @@ class AnswerConfigBlock extends React.Component {
 
     log('[moveRow]: ', from, to);
 
-    const update = swap(rows, from, to);
+    const movedRow = rows.find((row, index) => index === from);
+    const remainingRows = rows.filter((row, index) => index !== from);
```
andreeapescar, line 100: Can we simplify this? Use reduce

PR #745, opened 2021-10-29
Feat/pie print

PatriciaRomaniuc:


packages/explicit-constructed-response/src/main.js
```diff
@@ -54,29 +61,51 @@ export class Main extends React.Component {
...
+      note,
+      showNote,
+      env,
+      animationsDisabled,
+      alwaysShowCorrect,
+      maxLengthPerChoice
+    } = this.props;
     const { role } = env || {};
     const displayNote = (showCorrectAnswer || mode === 'view' && role === 'instructor') && showNote && note;
+    const mainClasses = alwaysShowCorrect ? classNames(classes.mainContainer, classes.noBorderColor)
```
andreeapescar, line 79: You could also use this format: `classNames('foo', { bar: true }); // => 'foo bar'`

andreeapescar (approved):

PR #761, opened 2021-11-18
feat(extended-text-entry):  Added the ability for instructor to annotate student's response with comments PD-744

andreeimiron:
https://illuminate.atlassian.net/browse/PD-744

A small demo with the new feature:

https://user-images.githubusercontent.com/63587739/142632389-a4834682-fd43-44f4-a042-071088e912bc.mov

Note: To select a word you can simply double click the word and the word will be selected.


packages/extended-text-entry/package.json
```diff
@@ -20,7 +20,8 @@
     "lodash.throttle": "^4.1.1",
     "prop-types": "^15.6.1",
     "react": "^16.8.1",
-    "react-dom": "^16.8.1"
+    "react-dom": "^16.8.1",
+    "uuid": "^8.3.2"
```
andreeapescar, line 24: My understanding is that you added this to generate uniq ids for the selections. Do you think we need to have a library for this, ok could we use a concatenation of itemId + new Date().getTime()?

packages/pie-models/src/pie/extended-text-entry/index.ts
```diff
@@ -78,6 +95,9 @@ export interface ExtendedTextEntryPie extends PieModel {
    * @default: 'bottom'
    */
   toolbarEditorPosition?: 'bottom' | 'top';
+
+  /** Indicates the predefined annotations */
```
andreeapescar, line 99: Maybe add few more details? Like what the predefined annotations are used for?

andreeapescar:
Very good job! ⭐

packages/extended-text-entry/package.json
andreeimiron: Not really. I removed the library and created a custom id based on start and end position of a annotation and new Date().getTime()

PR #794, opened 2021-12-23
chore(number line): remove snapPerTick and tickLabelOverrides from model

CarlaCostea:
- this are old properties not implemented and not wanted in pie


andreeapescar:
Andrei M will implement tickLabelOverrides. So we can keep them.

PR #799, opened 2022-01-06
feat(number-line): Added support for fractions PD-472

andreeimiron:
https://illuminate.atlassian.net/browse/PD-472

### DIFF
--- a/packages/number-line/configure/src/index.js
+++ b/packages/number-line/configure/src/index.js
@@ -3,12 +3,32 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { ModelUpdatedEvent } from '@pie-framework/pie-configure-events';
 import * as defaults from './defaults';
+import * as math from "mathjs";
 
 export default class NumberLine extends HTMLElement {
-  static createDefaultModel = (model = {}) => ({
-    ...defaults.model,
-    ...model
-  });
+  static createDefaultModel = (model = {}) => {
+    const normalizedModel = {
+      ...defaults.model,
+      ...model
+    };
+    const { graph: { labelStep, ticks: { minor, major } = {}} = {}} = normalizedModel;
+
+    if (labelStep && typeof labelStep === 'string') {
+      normalizedModel.graph.fraction = true;
+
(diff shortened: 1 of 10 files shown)


packages/number-line/configure/src/index.js
```diff
@@ -3,12 +3,32 @@ import React from 'react';
...
-    ...model
-  });
+  static createDefaultModel = (model = {}) => {
+    const normalizedModel = {
+      ...defaults.model,
+      ...model
+    };
+    const { graph: { labelStep, ticks: { minor, major } = {}} = {}} = normalizedModel;
+
+    if (labelStep && typeof labelStep === 'string') {
```
andreeapescar, line 16: Here is not enough to check if it's string. Check if it has a fraction format.

packages/number-line/controller/src/index.js
```diff
@@ -226,15 +227,28 @@ export function model(question, session, env) {
 
     if (graph) {
       const evaluateMode = env.mode === 'evaluate';
+      const { exhibitOnly, labelStep, ticks: { minor, major } = {}} = graph;
+
+      if (labelStep && typeof labelStep === 'string') {
```
andreeapescar, line 232: Same

packages/number-line/src/number-line/graph/ticks.jsx
```diff
@@ -79,13 +87,25 @@ export class Tick extends React.Component {
           x1="0.5"
           x2="0.5"
         />
+
```
andreeapescar, line 90: Make sure that if font size changes this approach is still looking as expected.

packages/number-line/configure/src/index.js
```diff
@@ -3,12 +3,32 @@ import React from 'react';
...
+      normalizedModel.graph.fraction = true;
+
+      // update the ticks frequency and label value to match the label step if needed
```
andreeapescar, line 19: Also, maybe you can extract this into a function? Same in controller. And add a comment like:
 // this function is duplicated in controller; at some point, use the same shared function

andreeapescar (changes requested):

PR #837, opened 2022-02-07
fix/PD-1581

CarlaCostea:
https://illuminate.atlassian.net/browse/PD-1581


packages/math-inline/src/simple-question-block.jsx
```diff
@@ -106,7 +105,7 @@ const SimpleQuestionBlock = withStyles(theme => ({
     color: color.text(),
     background: color.background(),
     border: `1px solid ${color.primaryLight()}`,
-    width: '100%',
+    width: 'fit-content',
```
andreeapescar, line 108: Have you checked this on multiple screen sizes?

CarlaCostea: yes, it looks good

PR #862, opened 2022-03-03
fix(hotspot): add support for --pie-zoom css property PD-360

sergiulucaci:
https://illuminate.atlassian.net/browse/PD-360

I will have to take care of drawing response also. Same issue happens there. 

I guess that will be another ticket - there will be more work than it is in this PR, wdyt @andreeapescar?

### DIFF
--- a/packages/hotspot/src/hotspot/index.jsx
+++ b/packages/hotspot/src/hotspot/index.jsx
@@ -27,9 +27,12 @@ class HotspotComponent extends React.Component {
 
         const zoomParsed = zoom?.replace(/\s/g, '');
         if (zoomParsed) {
-          this.setState({
-            scale: parseFloat(zoomParsed) / 100,
-          })
+          const newScale = parseFloat(zoomParsed) / 100;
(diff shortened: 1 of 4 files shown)


andreeapescar:
Yes, it would require another ticket. You can create one if you want to, and then check with George for ticket priority, @sergiulucaci

packages/hotspot/src/hotspot/index.jsx
```diff
@@ -2,17 +2,48 @@ import React from 'react';
...
+      mutations.forEach(() => {
+        const target = document.getElementById('question-container')?.style?.cssText
+        const zoom = target?.substring(
+          target.indexOf('--pie-zoom') + 11,
+          target.lastIndexOf('%')
+        );
+
+        const zoomParsed = zoom?.replace(/\s/g, '');
+        if (zoomParsed) {
+          this.setState({
```
andreeapescar, line 30: Do you think it would make sense to check if the value is different before updating the state?
Just to make sure we don't cal setState if it's not necessary?

PR #901, opened 2022-04-04
feat(inline-dropdown): added configurable validation PD-1698

andreeimiron:
https://illuminate.atlassian.net/browse/PD-1698


andreeapescar:
@pie-element/inline-dropdown@3.10.6-next.3

andreeapescar:
@pie-element/hotspot@4.10.16-next.1

PR #902, opened 2022-04-04
feat(select-text): added configurable validation PD-1700

andreeimiron:
https://illuminate.atlassian.net/browse/PD-1700


andreeapescar:
@pie-element/select-text@6.5.26-next.2

PR #903, opened 2022-04-05
feat(hotspot): added configurable validation PD-1702

andreeimiron:
https://illuminate.atlassian.net/browse/PD-1702


andreeapescar:
@pie-element/hotspot@4.10.16-next.1

PR #924, opened 2022-04-18
feat(graphing): Changed the consequences of toggling tools on or off PD-1608

andreeimiron:
https://illuminate.atlassian.net/browse/PD-1608


packages/graphing/configure/src/correct-response.jsx
```diff
@@ -91,32 +113,88 @@ export class CorrectResponse extends React.Component {
...
 
     set(model, `answers.${key}.marks`, marks);
     onChange(model);
   };
 
+  filterMarks = (tool) => {
+    const { model: { answers }} = this.props;
+    const filteredAnswers = {};
+
+    Object.entries(answers || {}).forEach(([key, answer]) => {
```
andreeapescar, line 138: Can we use a reduce here?

andreeimiron: sure

PR #939, opened 2022-05-05
feat(charting): create validate method in controller PD-1749

PatriciaRomaniuc:
This is marked as draft since there will need to be validation ui/ux implemented
https://illuminate.atlassian.net/browse/PD-1749


PatriciaRomaniuc:
Feature implemented in: https://github.com/pie-framework/pie-elements/pull/1153

PR #1096, opened 2022-10-12
feat: enable spanish and special characters in author PD-2120

andreeapescar:


andreeapescar:
https://illuminate.atlassian.net/browse/PD-2120

PR #1137, opened 2022-10-31
fix(drawing-response): for black, text color is changed to white-PD-2047

arimieandreea:


arimieandreea:
Jira link: https://illuminate.atlassian.net/browse/PD-2047

packages/drawing-response/src/drawing-response/drawable-palette.jsx
```diff
@@ -95,6 +99,9 @@ const styles = theme => ({
     marginRight: theme.spacing.unit * 2,
     marginTop: theme.spacing.unit * 2
   },
+  blackColorItem: {
+    color:'#FFFFFF',
```
PatriciaRomaniuc, line 103: Check if there is any color constant for this, don't use hex code.
color.text() should be it.

packages/drawing-response/src/drawing-response/drawable-palette.jsx
```diff
@@ -4,6 +4,7 @@ import { InputContainer } from '@pie-lib/render-ui';
 import Select from '@material-ui/core/Select';
 import MenuItem from '@material-ui/core/MenuItem';
 import { withStyles } from '@material-ui/core/styles';
+import classnames from 'classnames';
```
PatriciaRomaniuc, line 7: Just to check, it might already be implemented correctly: 
Always make sure that what new imports are used are also present in dependencies in package.json. Even though it works locally, the build won't work unless all the libraries used are explicitly declared in package.json.

packages/drawing-response/src/drawing-response/drawable-palette.jsx
```diff
@@ -45,7 +46,10 @@ class Palette extends React.Component {
               <MenuItem
                 key={value}
                 value={value}
-                className={classes.item}
+                className={classnames(classes.item, {
+                  [classes.blackColorItem]:
+                  value==='black'
```
PatriciaRomaniuc, line 51: add some spacing here

PatriciaRomaniuc (changes requested):

andreeimiron (approved):

PR #1153, opened 2022-11-08
feat(charting): UI/UX for validation PD-1750

CarlaCostea:
https://illuminate.atlassian.net/browse/PD-1750


packages/charting/configure/src/correct-response.jsx
```diff
@@ -216,14 +221,19 @@ export class CorrectResponse extends React.Component {
...
     const { categories } = this.state;
 
     return (
       <div>
         Define Correct Response
         <div className={classes.container}>
           <div className={classes.column} key="graph">
+            {correctAnswerErrors && (
+              <Typography component="div" type="body1" className={classes.noCorrectResponse}>
+              <span>{correctAnswerErrors.indenticalError || correctAnswerErrors.categoriesError}</span>
```
PatriciaRomaniuc, line 234: Can't indenticalError and categoriesError be defined at the same time? If so, shouldn't we display both errors?

CarlaCostea: no, if we get a categoriesError, then we won't get an error for identicalError

PR #1154, opened 2022-11-09
feat(categorize)/ changes to Remove After Placing controls and behavi…

arimieandreea:
…ors-PD-1974

### DIFF
--- a/packages/categorize/configure/src/defaults.js
+++ b/packages/categorize/configure/src/defaults.js
@@ -5,6 +5,7 @@ export default {
     choicesLabel: '',
     lockChoiceOrder: true,
     removeTilesAfterPlacing: false,
+    allowMultiplePlacementsEnabled: 'Yes',
     categoriesPerRow: 2,
     categories: [],
     alternates: [],
@@ -32,6 +33,10 @@ export default {
     lockChoiceOrder: {
       settings: false,
       label: 'Lock Choice Order'
+    },
+    allowMultiplePlacements: {
+      settings: true,
+      label: 'Allow Multiple Placements'
     },
     partialScoring: {
       settings: false,
--- a/packages/categorize/configure/src/design/choices/choice.jsx
+++ b/packages/categorize/configure/src/design/choices/choice.jsx
@@ -25,6 +25,7 @@ const canDrag = props => {
 
 export class Choice extends React.Component {
   static propTypes = {
+    allowMultiplePlacements:PropTypes.string,
     classes: PropTypes.object.isRequired,
     className: PropTypes.string,
     choice: PropTypes.object.isRequired,
@@ -63,8 +64,16 @@ export class Choice extends React.Component {
     onChange(choice);
   };
 
+  isCheckboxShown = (allowMultiplePlacements) => {
+    if(allowMultiplePlacements === 'Yes' || allowMultiplePlacements === 'No'){
+      return false;
+    }
+    return true;
+  };
+
   render() {
     const {
+      allowMultiplePlacements,
       classes,
       className,
       choice,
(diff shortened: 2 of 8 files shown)


arimieandreea:
Jira link: https://illuminate.atlassian.net/browse/PD-1974

packages/categorize/configure/src/defaults.js
```diff
@@ -5,6 +5,7 @@ export default {
     choicesLabel: '',
     lockChoiceOrder: true,
     removeTilesAfterPlacing: false,
+    allowMultiplePlacementsEnabled: 'Yes',
```
PatriciaRomaniuc, line 8: Does this property have any other value other than "Yes" or "No"? If not, please make it a boolean value

packages/categorize/configure/src/design/index.jsx
```diff
@@ -136,6 +136,18 @@ export class Design extends React.Component {
     return countInAnswer(choice.id, model.correctResponse);
   };
 
+  checkAllowMultiplePlacements = (allowMultiplePlacements, c) => {
+    if(allowMultiplePlacements === 'Yes'){
```
PatriciaRomaniuc, line 140: using a boolean value rather than string will make this check more simpe

packages/categorize/configure/src/design/index.jsx
```diff
@@ -136,6 +136,18 @@ export class Design extends React.Component {
...
+      return 0;
+    }
+    if(allowMultiplePlacements === 'No'){
```
PatriciaRomaniuc, line 143: add spacing

packages/categorize/configure/src/design/choices/choice.jsx
```diff
@@ -25,6 +25,7 @@ const canDrag = props => {
 
 export class Choice extends React.Component {
   static propTypes = {
+    allowMultiplePlacements:PropTypes.string,
```
PatriciaRomaniuc, line 28: add spacing

PatriciaRomaniuc (changes requested):

packages/categorize/configure/src/defaults.js
arimieandreea: Yes, it have another one 'Set per Choice'.

andreeimiron (commented):
You need to remove unnecessary spaces and add some spacing for better readability

packages/categorize/configure/src/design/choices/choice.jsx
```diff
@@ -63,8 +64,13 @@ export class Choice extends React.Component {
     onChange(choice);
   };
 
+  isCheckboxShown = allowMultiplePlacements => {
+    return !(allowMultiplePlacements === 'Yes' || allowMultiplePlacements === 'No');
```
PatriciaRomaniuc, line 68: Wouldn't this be more simple if you check allowMultiplePlacements === 'Set Per Choice' ? 
Also, maybe you can create a constant with these values 'Yes', 'No', 'Set Per Choice' since they are used multiple times. Something like
 `multiplePlacements: { enabled: 'Yes', disabled: 'No', perChoice: 'Set Per Choice' }`

packages/categorize/configure/src/design/index.jsx
```diff
@@ -136,6 +136,17 @@ export class Design extends React.Component {
...
+  checkAllowMultiplePlacements = (allowMultiplePlacements, c) => {
+    if (allowMultiplePlacements === 'Yes') {
+      return 0;
+    }
+    if (allowMultiplePlacements === 'No') {
+      return 1;
+    }
+    return c.categoryCount || 0;
+
+  };
```
PatriciaRomaniuc, line 148: This can be simplified by checking the 'Set per choice' value first. Something like this: 
`if (allowMultiplePlacements === 'Set Per Choice') return c.categoryCount || 0; 
return allowMultiplePlacements === 'No'`

PatriciaRomaniuc (changes requested):

PatriciaRomaniuc (approved):

PR #1173, opened 2022-11-16
Feat/pd 1980

arimieandreea:


andreeimiron (changes requested):
The code looks good, but you need to update the documentation with the new property

andreeimiron (approved):

PR #1178, opened 2022-11-18
fix(categorize) show alternate correct note if there is alternateResp…

arimieandreea:
…onse-PD-1978

### DIFF
--- a/packages/categorize/src/categorize/index.jsx
+++ b/packages/categorize/src/categorize/index.jsx
@@ -116,7 +116,7 @@ export class Categorize extends React.Component {
     return flexDirection;
   };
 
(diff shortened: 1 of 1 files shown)


arimieandreea:
Jira link: https://illuminate.atlassian.net/browse/PD-1978

packages/categorize/src/categorize/index.jsx
```diff
@@ -116,6 +116,8 @@ export class Categorize extends React.Component {
     return flexDirection;
   };
 
+  existAlternateResponse = (correctResponse) => correctResponse?.some(correctRes => correctRes.alternateResponses?.length >0);
```
PatriciaRomaniuc, line 119: Code looks good but spacing is a bit off here, maybe we can check if you have different a prettier setup.
There should always be spacing between operators.

andreeimiron (approved):

PR #1179, opened 2022-11-22
feat(categorize): categories per accept values up to 6-PD-1983

arimieandreea:


arimieandreea:
The changes from categories.jsx are done to avoid the warning regarding the unique key in items in case of adding divs

PatriciaRomaniuc (approved):

packages/categorize/configure/src/design/index.jsx
```diff
@@ -239,8 +240,8 @@ export class Design extends React.Component {
                   categoriesPerRow: categoriesPerRow.settings && numberField(categoriesPerRow.label,
                     {
                           label: categoriesPerRow.label,
-                          min: 1,
-                          max: 4,
+                          min: lowerMaximumValueCategoriesPerRow,
```
andreeimiron, line 243: the name is too long for this one. I think you can rename to `minCategoriesPerChoice` if I understand correctly what it refers to

andreeimiron (changes requested):
You need to add the new config property in the documentation

packages/pie-models/src/pie/categorize/index.ts
```diff
@@ -201,6 +201,11 @@ export interface CategorizeConfigure
    */
   maxCategories?: number;
 
+  /**
+   *  Lower maximum value for categories per row
```
PatriciaRomaniuc, line 205: Don't you mean minimum value? If that is what it represents

andreeimiron (approved):

PR #1292, opened 2023-02-08
fix(categorize): avoid long text overflow container-PD-2498

arimieandreea:


packages/categorize/configure/src/design/categories/__tests__/__snapshots__/choice-preview.test.jsx.snap
```diff
@@ -14,6 +14,7 @@ exports[`ChoicePreview snapshot renders 1`] = `
     onRemoveChoice={[Function]}
   >
     <HtmlAndMath
+      className="undefined"
```
CarlaCostea, line 17: Is there any issue with the className being undefined?

CarlaCostea (commented):
Is there any issue with the className being undefined?

packages/categorize/configure/src/design/categories/__tests__/__snapshots__/choice-preview.test.jsx.snap
arimieandreea: I think no. All tests have passed. I don't know exactly how snapshots should be updated in this case, but we have this in multiple-choice and match-list too

packages/categorize/configure/src/design/categories/choice-preview.jsx
```diff
@@ -27,7 +27,7 @@ export class ChoicePreview extends React.Component {
       <div className={classNames(classes.choicePreview, className)}>
         {choice ? (
           <Choice alternateResponseIndex={alternateResponseIndex} category={category} choice={choice} choiceIndex={choiceIndex} onRemoveChoice={() => this.delete()}>
-            <HtmlAndMath html={choice?.content} />
+            <HtmlAndMath html={choice?.content} className={`${classes.breakWord}`} />
```
andreeapescar, line 30: Why do you need the `${...}`? Why isn't it 
className={classes.breakWord} ?

arimieandreea: @andreeapescar if we send it as classes.breakWord, it will send it as an object and we need the  name of the classname (string) as props

PR #1294, opened 2023-02-09
feat: remove force enabled logic from withRubric property PD-2624

PatriciaRomaniuc:
https://illuminate.atlassian.net/browse/PD-2624


andreeapescar:
@PatriciaRomaniuc can you please add a new commit with Breaking Changes? Even if we'll move this code on the player side, we have to make sure nothing breaks if someone uses this as it was - with forceEnabled (I know, probably nobody, but still...)

PR #1350, opened 2023-03-15
fix(grapnig): graphDimensions:settings setting does not behave as exp…

arimieandreea:
…ected

### DIFF
--- a/packages/graphing/configure/src/defaults.js
+++ b/packages/graphing/configure/src/defaults.js
@@ -32,7 +32,6 @@ export default {
     includeAxes: true,
     labels: {},
     labelsEnabled: true,
-    dimensionsEnabled: true,
     padding: true,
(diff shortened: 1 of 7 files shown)


packages/graphing/configure/src/defaults.js
```diff
@@ -32,6 +32,7 @@ export default {
     includeAxes: true,
     labels: {},
     labelsEnabled: true,
+    dimensionsEnabled: true,
```
andreeimiron, line 35: I think you should remove this from defaults and move the logic inside componentDidMount or constructor. I suggest to check if this value is set to the model and if not, assign the value from the config defaults.

andreeimiron (changes requested):

andreeimiron (approved):

PR #1367, opened 2023-03-28
feat(passage): make paired passage tabs accessible to keyboard users PD-2445

CarlaCostea:
https://illuminate.atlassian.net/browse/PD-2445


packages/passage/src/stimulus-tabs.jsx
```diff
@@ -75,6 +75,37 @@ class StimulusTabs extends React.Component {
...
+    } else if (key === 'Enter' || key === ' ') {
+      event.preventDefault();
+      newTabIndex = tabs.findIndex((tab) => tab.id === currentTabId);
+    }
+  
+    if (newTabIndex !== -1) {
+      this.handleChange(event, tabs[newTabIndex].id);
+    }
+  };
+
```
andreeapescar, line 108: Couldn't we make this nicer with a switch? Also, it looks like we call event.preventDefault every time one if is called. Can we only call it if newTabIndex !== -1?

CarlaCostea: sure, I added the switch and refactored the function.

PR #1398, opened 2023-05-01
fix(charting): if correctAnswer is not defined in the model set correctAnswer to equal the data PD-2626

CarlaCostea:
https://illuminate.atlassian.net/browse/PD-2626


packages/charting/configure/src/correct-response.jsx
```diff
@@ -235,7 +237,7 @@ export class CorrectResponse extends React.Component {
                 domain={domain}
                 range={range}
                 charts={charts}
-                data={categories}
+                data={correctResponseCategories}
```
andreeapescar, line 240: This only renders a correct response, but I don't think this sets the correct response on the model. So let's make sure we have that.

PR #1425, opened 2023-05-31
fix(multiple-choice): avoid duplicate id for choice in case ebsr-PD-2831

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/jira/software/c/projects/PD/issues/PD-2831

PR #1434, opened 2023-06-09
fix(graphing): PD-2873 alternates out of order when deleted

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/jira/software/c/projects/PD/issues/PD-2873

PR #1457, opened 2023-07-10
fix(graphing): PD-2976 - add validation info to correct answer

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/jira/software/c/projects/PD/issues/PD-2976

PR #1460, opened 2023-07-12
Fix/pd 2904

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PD-2904

PR #1477, opened 2023-08-11
feat(select-text): design updates in Evaluate

arimieandreea:


packages/select-text/controller/src/index.js
```diff
@@ -6,10 +6,29 @@ import defaults from './defaults';
 
 const log = debug('@pie-element:select-text:controller');
 
-const buildTokens = (tokens, evaluateMode) => {
-  tokens = tokens || [];
+const isAnswerSelected = (token, selectedToken) => {
+  return (
+    token.correct === true &&
```
andreeapescar, line 11: Please add sanity check on token, and remove { return (...) }. You can simply do => (...)

packages/select-text/controller/src/index.js
```diff
@@ -118,7 +137,7 @@ export const model = (question, session, env) => {
   return new Promise((resolve) => {
     log('[model]', 'normalizedQuestion: ', normalizedQuestion);
     log('[model]', 'session: ', session);
-    const tokens = buildTokens(normalizedQuestion.tokens, env.mode === 'evaluate');
+    const tokens = buildTokens(normalizedQuestion.tokens, session.selectedTokens, env.mode === 'evaluate');
```
andreeapescar, line 140: Add sanity check on session

PR #1492, opened 2023-08-25
fix(passage): update font-size + refactoring PD-3133

andreeimiron:
https://illuminate.atlassian.net/browse/PD-3133

### DIFF
--- a/packages/passage/src/stimulus-tabs.jsx
+++ b/packages/passage/src/stimulus-tabs.jsx
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import Tabs from '@material-ui/core/Tabs';
 import Tab from '@material-ui/core/Tab';
 import { withStyles } from '@material-ui/core/styles';
-import Typography from '@material-ui/core/Typography';
 import { color, Purpose } from '@pie-lib/render-ui';
 import classNames from 'classnames';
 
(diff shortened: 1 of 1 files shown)


arimieandreea (approved):

packages/passage/src/stimulus-tabs.jsx
```diff
@@ -15,11 +14,10 @@ const styles = (theme) => ({
   },
   tab: {
     backgroundColor: color.background(),
-    fontSize: '0.8125em',
+    fontSize: theme.typography.fontSize,
```
arimieandreea, line 17: please test what happens if font-size on html is 10px, and on body is14px. I have the impression that it will not change accordingly to this both. Test both for tabs an text

packages/passage/src/stimulus-tabs.jsx
```diff
@@ -33,37 +31,16 @@ const styles = (theme) => ({
   buttonContainer: {
     background: color.background(),
   },
+  tabContainer: {
+    backgroundColor: color.background(),
+    color: color.text(),
+    padding: theme.spacing.unit * 2,
```
arimieandreea, line 37: have you had in mind that padding was different based on props.multiple?

arimieandreea (changes requested):

packages/passage/src/stimulus-tabs.jsx
andreeimiron: I changed the padding in order to align with the last UI changes on authoring side

arimieandreea (approved):

PR #1501, opened 2023-09-05
Feat/pd 599

arimieandreea:


packages/hotspot/configure/src/hotspot-polygon.jsx
```diff
@@ -1,24 +1,37 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Group, Line } from 'react-konva';
+import { Group, Line, Circle } from 'react-konva';
 import { withStyles } from '@material-ui/core/styles/index';
 
 class PolComponent extends React.Component {
-  static getDerivedStateFromProps(nextProps) {
-    const { points } = nextProps;
+  constructor(props) {
```
andreeapescar, line 7: Do you actually need this constructor?

packages/hotspot/configure/src/hotspot-polygon.jsx
```diff
@@ -50,57 +66,100 @@ class PolComponent extends React.Component {
...
+    }
   };
 
   render() {
     const { classes, correct, hotspotColor, outlineColor, strokeWidth = 5 } = this.props;
-    const { points, x, y } = this.state;
+    const { points, x, y, hovered } = this.state;
+
+    const calculatedStrokeWidth = correct ? strokeWidth : hovered ? 1 : 0;
+    const calculatedStroke = correct ? outlineColor : hovered ? '#00BFFF' : '';
```
andreeapescar, line 123: Can you please store this color code somewhere? If we don't have it in theme, maybe at least declare a const for it?

packages/hotspot/configure/src/hotspot-rectangle.jsx
```diff
@@ -30,13 +43,36 @@ class RectComponent extends React.Component {
     });
   };
 
+  onResizeEnd = () => {
+    const { onDragEnd, id } = this.props;
+    // transformer is changing scale of the node
```
andreeapescar, line 48: love the fact that you added comments 🙏

PR #1542, opened 2023-10-19
fix(match): Introduced  function to ensure row IDs are sequential and unique. PD-3234

CarlaCostea:
https://illuminate.atlassian.net/browse/PD-3234

Add row ID validation to Configure component

- Introduced `validateRowsID` function to ensure row IDs are sequential and unique.
- Implemented ID validation logic in `componentDidMount` to correct any inconsistencies in the row IDs upon component mount.

Refactor rowIdCounter initialization to ensure highest id value

- Calculate maxId from all rows instead of relying on the last row's id.
- Add checks to handle potential cases where props.model.rows is undefined or empty.
- Set rowIdCounter to maxId + 1, ensuring a unique and sequential id.

This addition ensures that row IDs in the Configure component are always consistent and avoids potential issues with duplicate or missing IDs.

### DIFF
--- a/packages/match/configure/src/configure.jsx
+++ b/packages/match/configure/src/configure.jsx
@@ -54,6 +54,31 @@ class Configure extends React.Component {
     };
   }
 
+  componentDidMount() {
+    const { updatedRows, wasChanged } = this.validateRowsID(this.props.model.rows);
+    
+    if (wasChanged) {
+      const newModel = { ...this.props.model, rows: updatedRows };
+      this.props.onModelChanged(newModel);
+    }
+  }
+
+  validateRowsID(rows) {
+    let wasChanged = false;
+
+    const updatedRows = (rows || []).map((row, index) => {
+      if (row.id !== index + 1) {
+        wasChanged = true;
+
+        return { ...row, id: index + 1 };
+      }
+
+      return row;
+    });
+
+    return { updatedRows, wasChanged };
+  }
+
   onTabChange = (event, value) => {
     this.setState({ activeTab: value });
   };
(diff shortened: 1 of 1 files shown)


packages/match/configure/src/configure.jsx
```diff
@@ -54,6 +54,31 @@ class Configure extends React.Component {
...
+
+        return { ...row, id: index + 1 };
+      }
+
+      return row;
+    });
+
+    return { updatedRows, wasChanged };
+  }
+
```
PatriciaRomaniuc, line 81: Does this assure that if there are many changes made to the questions (some questions are deleted, added, deleted again) the model will have correct id's and will evaluate correctly? I am asking because it's a bit difficult to verify this from just the code.

CarlaCostea: @PatriciaRomaniuc No, this code mainly serves as a validation to confirm that each row's id in the model given by the user is distinct. I was under the impression that the model was directly supplied by the user. I haven't been able to replicate the error when creating an item. if there are many changes made to the questions (some questions are deleted, added, deleted again) the model will have correct id's and will evaluate correctly. Have I overlooked something?

CarlaCostea: @PatriciaRomaniuc as we discussed I added the following changes:

Refactor `rowIdCounter` initialization to ensure highest id value

- Calculate maxId from all rows instead of relying on the last row's id.
- Add checks to handle potential cases where `props.model.rows` is undefined or empty.
- Set `rowIdCounter` to maxId + 1, ensuring a unique and sequential id.

PatriciaRomaniuc (approved):

andreeimiron (approved):

PR #1599, opened 2023-12-15
fix(charting): fix state sync issues in Define Correct Response Chart when Define Chart data is modified or removed PD-3164

CarlaCostea:
https://illuminate.atlassian.net/browse/PD-3164

### DIFF
--- a/packages/charting/configure/src/correct-response.jsx
+++ b/packages/charting/configure/src/correct-response.jsx
@@ -99,9 +99,9 @@ const removeCategory = (correctAnswer, data, positionToRemove) => {
 };
 
 export const getUpdatedCategories = (nextProps, prevProps, prevState) => {
-  const nextData = nextProps.model.data || [];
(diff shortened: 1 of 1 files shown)


packages/charting/configure/src/correct-response.jsx
```diff
@@ -97,6 +98,57 @@ const removeCategory = (correctAnswer, data, positionToRemove) => {
   return addCategoryProps(correctAnswerData, data);
 };
 
+export const getUpdatedCategories = (nextProps, prevProps, prevState) => {
+  const nextData = nextProps.model.data || [];
```
andreeapescar, line 102: Everything looks good, but I will insist with sanity check.
Everything that has \. something, has to be checked.
You either make sure nextProps is not null/undefined (if makes sense in your code), like : nextProps = {...nextProps}
Or you check it const nextData = (nextProps && nextProps.model && nextProps.model.data) || [];

PR #1633, opened 2024-01-10
fix(number-line): Fix issue with domain and ticks as per SC-28651

akshayhiwale:
Fixed number line issue as per SC-28651/SC-28674
1. Fix when domain min and max are same screen gets blank
2. Added min and max for number text field.
3. Ticks model not getting set after saving item.
4. Changed default tickintervaltype to decimal to accommodate legacy NL items.
5. AvailableTypes not getting loaded correctly after saving Item

### DIFF
--- a/packages/number-line/configure/src/domain.jsx
+++ b/packages/number-line/configure/src/domain.jsx
@@ -23,7 +23,13 @@ export class Domain extends React.Component {
 
   change(key, event, value) {
     const { onChange } = this.props;
-    const update = { ...this.props.domain, [key]: value };
+    let update;
+    //Added condition when min and max is same, then it should not update the value
+    if ((key === 'min' && value === this.props.domain.max) || (key === 'max' && value === this.props.domain.min)) {
+      update = { ...this.props.domain };
+    } else {
+      update = { ...this.props.domain, [key]: value };
+    }
     onChange(sort(update));
   }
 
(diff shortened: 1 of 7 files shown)


packages/number-line/configure/src/domain.jsx
```diff
@@ -23,7 +23,13 @@ export class Domain extends React.Component {
 
   change(key, event, value) {
     const { onChange } = this.props;
-    const update = { ...this.props.domain, [key]: value };
+    let update;
+    //Added condition when min and max is same, then it should not update the value
```
andreeimiron, line 27: in this case isn't it better to just return and not call onChange?

```suggestion

if ((key === 'min' && value === this.props.domain.max) || (key === 'max' && value === this.props.domain.min)) {
return;
}

const update = { ...this.props.domain, [key]: value };
onChange(sort(update));
```

andreeimiron (changes requested):

packages/number-line/configure/src/domain.jsx
akshayhiwale: If we just return, then the new value in the input field remain as it is. It does not get revert back to old value. Thats why calling onChange().

packages/number-line/configure/src/main.jsx
```diff
@@ -539,9 +567,9 @@ export class Main extends React.Component {
           {model.widthEnabled && (
             <Size
               size={graph}
-              min={numberLineDimensions.min}
-              max={numberLineDimensions.max}
-              step={numberLineDimensions.step}
+              min={numberLineDimensions ? numberLineDimensions.min : 400}
```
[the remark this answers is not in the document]
akshayhiwale: Fixed

andreeapescar (commented):
Awesome PR!

PR #1689, opened 2024-02-08
fix(categorize): no need to remove tags for preview prompt - PD-3440

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PD-3440

PR #1690, opened 2024-02-08
feat(number-line): Update Tick Interval logic for number line labels | SC-28865

akshayhiwale:
feat(number-line): Update Tick Interval logic for number line labels | SC-28865


andreeimiron (approved):

andreeapescar:
You have major changes of logic in this PR and this is really hard to follow...

I would highly appreciate if you could make explanatory video recordings, describing what you've changed (comparison with previous version) and how this is expected to work, it would be a lot easier to understand the code in the PR. Thank you. 

One side note, I see you have many "if - else if - else if - else" statements, maybe you can use "switch" instead?

PR #1699, opened 2024-02-09
fix(complex-rubric): add rubricless as default option PD-3451

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PD-3451

PR #1704, opened 2024-02-12
feat(categorize): ensure categoryCount is set PD-2974

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PD-2974

PR #1716, opened 2024-02-16
feat(hotspot): get rid of the authoring tooltip PD-3462

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/jira/software/c/projects/PD/issues/PD-3462

PR #1723, opened 2024-02-22
fix(multi-trait-rubric): scorePointsLabels shouve have length 1 or 1 …

arimieandreea:
…based on excludeZero PD-3468


arimieandreea:
https://illuminate.atlassian.net/jira/software/c/projects/PD/issues/PD-3468

PR #1729, opened 2024-02-29
fix(multi-trait-rubric): additional verif for empty string to avoid ad…

arimieandreea:
…ditional elemnts to be added PD-3492


arimieandreea:
https://illuminate.atlassian.net/browse/PD-3492

PR #1734, opened 2024-03-04
feat: Add configuration for custom buttons PD-3506

barta-adrian:
https://illuminate.atlassian.net/browse/PD-3506


andreeapescar (commented):
Looks great!

Number-line seems to have another input as well: https://examples.dev.pie-api.com/examples/pie-number-line.html

Do you think we should customize this one as well?

Screenshot 2024-03-04 at 12 41 13

andreeimiron (approved):

PR #1794, opened 2024-03-18
feat: Add configuration for custom buttons PD-3517

barta-adrian:
https://illuminate.atlassian.net/browse/PD-3517


andreeimiron (approved):

packages/drag-in-the-blank/configure/src/main.jsx
```diff
@@ -183,6 +185,15 @@ export class Main extends React.Component {
...
 
+    const getPluginProps = (props) => {
+      return Object.assign(
+        {
+          ...baseInputConfiguration,
+        },
+        props || {},
+      );
+    };
+
```
PatriciaRomaniuc, line 196: same here, you don't need Object.assign

PR #1870, opened 2024-04-15
Feat(rubric; multi-trait-rubric) : implement tests and adjustments in validateModel method

dianagociu:
https://illuminate.atlassian.net/browse/PD-3608

### DIFF
--- a/packages/multi-trait-rubric/configure/src/__tests__/index.test.jsx
+++ b/packages/multi-trait-rubric/configure/src/__tests__/index.test.jsx
@@ -10,311 +10,10 @@ jest.mock('react-dom', () => ({
 const model = (extras) => ({
     id: '1',
     element: 'multi-trait-rubric',
-    visibleToStudent: true,
-    halfScoring: false,
-    pointLabels: true,
-    description: false,
-    standards: false,
-    excludeZero: false,
-    addScaleEnabled: true,
-    maxPointsEnabled: true,
-    scales: [
-        {
-            maxPoints: 4,
-            scorePointsLabels: ['Non-Scorable', 'Developing', 'Progressing', 'Effective', 'Strong'],
-            traitLabel: 'Trait',
-            traits: [
-                {
-                    name: 'Ideas',
-                    standards: [],
-                    description: 'the main message',
-                    scorePointsDescriptors: [
-                        'Student’s response is blank, not in English, not legible, or does not respond to the prompt.',
-                        'Topic undefined and/or difficult to follow\n' + '\n' + 'Details are unclear',
-                        'Topic too broad\n' + '\n' + 'Details are limited',
-                        'Writing stays on topic\n' + '\n' + 'Complete details given',
-                        'Strong control of topic\n' + '\n' + 'Relevant, accurate, specific details that support topic',
-                    ],
-                },
-                {
-                    name: 'Organization',
-                    standards: [],
-                    description: 'the internal structure of the piece',
-                    scorePointsDescriptors: [
-                        'Student’s response is blank, not in English, not legible, or does not respond to the prompt.',
-                        'Does not have a beginning, middle and/or end\n' +
-                        '\n' +
-                        'Does not have a lead and/or conclusion\n' +
-                        '\n' +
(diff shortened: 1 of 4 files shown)


packages/multi-trait-rubric/configure/src/__tests__/index.test.jsx
```diff
@@ -0,0 +1,530 @@
...
+import defaults from '../defaults';
+import { ModelUpdatedEvent, InsertSoundEvent, DeleteSoundEvent } from '@pie-framework/pie-configure-events';
+import React from "react";
+
+jest.mock('react-dom', () => ({
+    render: jest.fn(),
+}));
+
+const model = (extras) => ({
+    id: '1',
```
andreeapescar, line 11: Let's reduce the amount of information here. You miht not need all the model properties.
Another suggestion is to add more clear labels, like "Trait 1", "Trait 2" etc.

packages/multi-trait-rubric/configure/src/__tests__/index.test.jsx
```diff
@@ -0,0 +1,530 @@
...
+                        'Handwriting poor\n' + '\n' + 'Overall appearance is distracting to unacceptable',
+                        'Handwriting is generally legible\n' + '\n' + 'Overall appearance is acceptable or better',
+                    ],
+                },
+            ],
+        },
+    ],
+    ...extras,
+});
+const configuration = () => ({
```
andreeapescar, line 191: Let's not copy defaults here. Instead, you can add a test that if you don't set any configuration, the defaults will be used & a test that if you add "SOME configuration properties, they will overwrite the defaults.

packages/multi-trait-rubric/configure/src/__tests__/index.test.jsx
```diff
@@ -0,0 +1,530 @@
...
+                                scorePointsDescriptors: ['Descriptor 1', 'Descriptor 2', 'Descriptor 3'],
+                            },
+                        ],
+                    },
+                ],
+            });
+            const validatedModel = element.updateModelAccordingToReceivedProps(invalidModel);
+            expect(validatedModel.scales[0].scorePointsLabels.length).toBe(3);
+            expect(validatedModel.scales[0].traits[0].scorePointsDescriptors.length).toBe(3);
+        });
```
andreeapescar, line 527: Add more tests based on content, instead of content length.

packages/multi-trait-rubric/configure/src/index.js
```diff
@@ -40,13 +46,15 @@ export default class MultiTraitRubricElement extends HTMLElement {
       }
 
       (scale.traits || []).forEach(trait => {
-        trait.scorePointsDescriptors = [ ...trait.scorePointsDescriptors ];
+        if (!trait) return;
```
andreeapescar, line 49: Let's make sure we actually update the model and make it valid, Returning undefined here will break the code.

packages/multi-trait-rubric/configure/src/index.js
```diff
@@ -16,22 +16,28 @@ export default class MultiTraitRubricElement extends HTMLElement {
...
+  updateModelAccordingToReceivedProps = (m) => {
+    if (!m) {
+      return { ...defaults.model };
+    }
+
+    const validatedModel = { ...m };
     const { scales, excludeZero } = validatedModel;
 
     (scales || []).forEach(scale => {
+      if (!scale) return;
```
andreeapescar, line 28: Let's make sure we actually update the model and make it valid, Returning undefined here will break the code.

packages/multi-trait-rubric/configure/src/index.js
```diff
@@ -16,22 +16,28 @@ export default class MultiTraitRubricElement extends HTMLElement {
     this._configuration = configurationWithDefaults();
   }
 
-  validateModel = (m) => {
-    const validatedModel = m;
+  updateModelAccordingToReceivedProps = (m) => {
+    if (!m) {
+      return { ...defaults.model };
```
andreeapescar, line 21: I would rather return existing model if there's no new model.

andreeapescar (approved):

PR #1878, opened 2024-04-17
feat(charting, graphing-solution-set): Charting Item Relative Scoring and Label Characters Limit SC-29795, SC-30682

akshayhiwale:
1. Implemented axis label character limit settings for Charting item
2. Added extra prop on outcome object for scores API for relative scoring
3. Punctuation fix in GSS item.
https://illuminate.atlassian.net/browse/SC-29795
https://illuminate.atlassian.net/browse/SC-30682

### DIFF
--- a/packages/charting/configure/src/charting-config.jsx
+++ b/packages/charting/configure/src/charting-config.jsx
@@ -128,6 +128,7 @@ export class ChartingConfig extends React.Component {
       authorNewCategoryDefaults,
       chartingOptions,
       mathMlOptions = {},
+      charactersLimit,
     } = this.props;
     const { dialog } = this.state;
     const { domain = {}, range = {} } = model || {};
@@ -167,6 +168,7 @@ export class ChartingConfig extends React.Component {
               labelsPlaceholders={labelsPlaceholders}
               titlePlaceholder={titlePlaceholder?.label}
               mathMlOptions={mathMlOptions}
+              charactersLimit={charactersLimit}
             />
             {model.changeAddCategoryEnabled && (
               <div>
--- a/packages/charting/configure/src/charting-config.jsx
+++ b/packages/charting/configure/src/charting-config.jsx
@@ -128,7 +128,7 @@ export class ChartingConfig extends React.Component {
       authorNewCategoryDefaults,
       chartingOptions,
       mathMlOptions = {},
-      charactersLimit,
+      labelsCharactersLimit,
     } = this.props;
     const { dialog } = this.state;
     const { domain = {}, range = {} } = model || {};
@@ -168,7 +168,7 @@ export class ChartingConfig extends React.Component {
               labelsPlaceholders={labelsPlaceholders}
               titlePlaceholder={titlePlaceholder?.label}
               mathMlOptions={mathMlOptions}
-              charactersLimit={charactersLimit}
+              labelsCharactersLimit={labelsCharactersLimit}
             />
             {model.changeAddCategoryEnabled && (
               <div>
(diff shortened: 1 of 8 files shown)


packages/charting/configure/src/charting-config.jsx
```diff
@@ -128,6 +128,7 @@ export class ChartingConfig extends React.Component {
       authorNewCategoryDefaults,
       chartingOptions,
       mathMlOptions = {},
+      charactersLimit,
```
andreeimiron, line 131: based on the request on pie-lib side, I suggest to rename this to `labesCharactersLimit`

packages/charting/controller/src/index.js
```diff
@@ -138,10 +151,16 @@ export const getScore = (question, session, env = {}) => {
     }
   }
 
-  return {
+  let score = {
```
andreeimiron, line 154: you can use `const` instead `let`

packages/charting/controller/src/index.js
```diff
@@ -222,10 +241,15 @@ export function model(question, session, env) {
 
 export function outcome(model, session, env) {
   return new Promise((resolve) => {
-    resolve({
-      score: getScore(model, session, env).score,
+    let scoreObject = getScore(model, session, env);
```
andreeimiron, line 244: you can use `const` instead `let`

packages/charting/controller/src/index.js
```diff
@@ -222,10 +241,15 @@ export function model(question, session, env) {
...
+    let result = {
```
andreeimiron, line 245: you can use `const` instead `let`

andreeimiron (changes requested):

packages/charting/controller/src/index.js
```diff
@@ -55,18 +56,25 @@ export const getScore = (question, session, env = {}) => {
...
 
     const scoreForLabelAndValueEditable = (answer, corrAnswer) => {
-      const { value, label } = answer;
-
+      const { value, label, index } = answer;
+      let firstEquality = false;
+      let secondEquality = false;
       maxScore += 2;
 
       if (value === corrAnswer.value) {
```
andreeapescar, line 64: It's a little bit confusing what firstEquality and secondEquality mean.

I assume:
firstEquality - if value is correct
secondEquality - if label is also correct
Is my assumption correct? If yes, can we maybe rename the variables, please?

Maybe we can store those values in variables and refactor where needed.
Something ike
`valueIsCorrect = value === corrAnswer.value`
`labelIsCorrect = checkLabelsEquality(label, corrAnswer.label)`

packages/charting/controller/src/index.js
```diff
@@ -129,6 +139,9 @@ export const getScore = (question, session, env = {}) => {
           result = 0;
           answers[index].correctness.label = 'incorrect';
         }
+        if (value === corrAnswer.value && lowerCase(label) === lowerCase(corrAnswer.label)) {
```
andreeapescar, line 142: Same here. I would store these equalities in a variable and use their value

andreeapescar (changes requested):

packages/charting/configure/src/charting-config.jsx
akshayhiwale: Done

packages/charting/controller/src/index.js
akshayhiwale: Done

packages/charting/controller/src/index.js
akshayhiwale: Done

packages/charting/controller/src/index.js
akshayhiwale: Done

packages/charting/controller/src/index.js
akshayhiwale: Done

packages/charting/controller/src/index.js
akshayhiwale: Done

PR #1898, opened 2024-05-03
fix(image-cloze-association): filter out only the first found respons…

arimieandreea:
…e PD-3718


arimieandreea:
https://illuminate.atlassian.net/browse/PD-3718

andreeimiron (approved):

PR #1924, opened 2024-05-29
Fix/pd 3738 reload graph controller

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PD-3738

PR #2000, opened 2024-07-05
fix(categorize): correct deletion of a choice from alternates, setMax…

arimieandreea:
…ChoicesPerCategory alternates PD-3895


arimieandreea:
https://illuminate.atlassian.net/browse/PD-3895

PR #2017, opened 2024-07-11
fix(graphing): key legend only in evaluate, adjust model function con…

arimieandreea:
…troller PD-3937


arimieandreea:
https://illuminate.atlassian.net/browse/PD-3937

dianagociu (approved):

PR #2020, opened 2024-07-11
feat(number-line): add Settings Panel and authoring view settings PD-3932

CarlaCostea:
https://illuminate.atlassian.net/browse/PD-3932


andreeapescar:
Can we add settingsPanelDisabled configuration here, in case SC doesn't want to show the Settings Panel?

PR #2044, opened 2024-07-19
fix(image-cloze-association): fixed controller duplicate methods in p…

PatriciaRomaniuc:
…layer to also use camelizeKeys method PD-3942

https://illuminate.atlassian.net/browse/PD-3942

### DIFF
--- a/packages/image-cloze-association/src/utils-correctness.js
+++ b/packages/image-cloze-association/src/utils-correctness.js
@@ -1,4 +1,7 @@
 // functions also used in controller/src/utils.js
+// camelize keys is needed to convert the keys from snake_case to camelCase
+// this is also done in the controller
+import { camelizeKeys } from 'humps';
 
 const getAllCorrectAnswers = (answers, responses) =>
   (answers || []).map((answer) => ({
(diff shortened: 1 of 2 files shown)


packages/image-cloze-association/src/utils-correctness.js
```diff
@@ -66,10 +70,11 @@ export const getUnansweredAnswers = (answers, validation) => {
 };
 
 export const getAnswersCorrectness = (answers, validation) => {
+  const camelizedValidation = camelizeKeys(validation);
```
andreeapescar, line 73: Could we use lodash camelCase instead of adding another library just for camel casing?
https://www.geeksforgeeks.org/lodash-_-camelcase-method/

PatriciaRomaniuc: I used the same library that is used on the controller side. But i can remove it from there too and change it to lodash

PatriciaRomaniuc: @andreeapescar I am not sure if it's actually better since I will have to recursively camel case the keys of the object. So write my own function both in controller and player and use that combined with lodash and not the one from humps. 
If you think that's better, I am open to it. 

The function is not difficult to write but it will be duplicate. Something like this:
```
const camelizeKeys = (obj) => {
  if (isArray(obj)) {
    return obj.map(camelizeKeys);
  }

  if (isObject(obj)) {
    return Object.entries(obj).reduce((acc, [key, value]) => {
      acc[camelCase(key)] = camelizeKeys(value);
      return acc;
    }, {});
  }

  return obj;
}; 
```

andreeapescar: Oh I see, I didn't realize we're already using it and that it works on objects. Thank you for the explanation. I can see it only has 10kb, so I guess it's ok to keep it, at least for now. Thank you.

PR #2065, opened 2024-07-30
feat: use a larger padding if there's enough room, use smaller paddin…

andreeapescar:
…g otherwise PD-3871


andreeapescar:
https://illuminate.atlassian.net/browse/PD-3871

PR #2078, opened 2024-08-12
feat(math-inline): Enhance keyboard navigation, focus and aria attributes for math editors PD-2453

CarlaCostea:
https://illuminate.atlassian.net/browse/PD-2453

- Improved keyboard navigation experience for math editor input fields.
- Ensured that on-screen math keypad opens and is accessible for keyboard-only users.
- Fixed focus issues for keyboard, touch, and click users.
- Addressed the issue where aria attributes were not updated when the response type changed.
- Updated instructions to reflect the new behavior of the math keypad, clarifying that it does not open automatically but can be accessed and navigated with keyboard input.
- Added instructions for keyboard users to use the down arrow key to "drill down" into the keypad and the escape key to "drill up" back to the input field.


packages/math-inline/src/main.jsx
```diff
@@ -290,7 +297,8 @@ export class Main extends React.Component {
                         instructionsElement = document.createElement('span');
                         instructionsElement.id = instructionsId;
                         instructionsElement.className = classes.srOnly;
-                        instructionsElement.textContent = 'This field automatically displays a math keypad. Both keypad and keyboard input are accepted, and keyboard entry accepts LaTeX markup.';
+                        instructionsElement.textContent =
+                        'This field supports both keypad and keyboard input. Use the keyboard to access and interact with the on-screen math keypad, which accepts LaTeX markup. Use the down arrow key to open the keypad and navigate its buttons. Use the escape key to close the keypad and return to the input field.';
```
PatriciaRomaniuc, line 301: I am wondering if we should use the translator for this since it's on the player side. Maybe it also needs to support spanish

packages/math-inline/src/main.jsx
```diff
@@ -299,17 +307,70 @@ export class Main extends React.Component {
...
+          });
+        }
+    
+        if (!isAnswerInputFocused && isClickOrTouchEvent) {
+          this.setState({ activeAnswerBlock: '' });
+        }
+    
+        if (event.key === 'Escape') {
+          this.setState({ activeAnswerBlock: '' });
+        }
```
PatriciaRomaniuc, line 343: This could be just one if statement since it has the same body
```
if ((!isAnswerInputFocused && isClickOrTouchEvent) || event.key === 'Escape') {
          this.setState({ activeAnswerBlock: '' });
        }
```

packages/math-inline/src/main.jsx
CarlaCostea: I've added this question in the feedback request for George: https://illuminate.atlassian.net/browse/PD-2453?focusedCommentId=1503466

PR #2131, opened 2024-09-10
feat(fraction-model): Implementation of new Fraction Model | SC-30716

akshayhiwale:
1. New Fraction Model Item type is implemented.
2. Test Cases are added.
3. Documentation is added.
4. Demo is added.
5. Snapshots are updated.
Confluence Link - https://illuminate.atlassian.net/wiki/spaces/SC/pages/17224564987/Create+new+Fraction+Model+Item+Type


packages/fraction-model/configure/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@pie-element/fraction-model-configure",
+  "private": true,
+  "version": "2.18.1",
```
andreeimiron, line 4: we should have an initial version like 1.0.0

packages/fraction-model/controller/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@pie-element/fraction-model-controller",
+  "private": true,
+  "version": "2.17.1",
```
andreeimiron, line 4: we should use an initial version like 1.0.0

packages/fraction-model/configure/src/model-options.jsx
```diff
@@ -0,0 +1,146 @@
...
+}
+
+const styles = () => ({
+  groupInline: {
+    alignItems: 'center',
+    display: 'flex',
+    gap: '1.25rem',
+  },
+  group: {
+    margin: '0.75rem 0',
```
andreeimiron, line 116: I suggest using 'px' instead of 'rem'. We had several issues caused by 'rem' and we try to avoid it.

packages/fraction-model/configure/src/number-text-field.jsx
```diff
@@ -0,0 +1,39 @@
...
+});
+
+const miniStyles = () => ({
+  textField: {
+    maxWidth: '120px',
+    width: '120px',
+    marginTop: '0',
+    '& [class^="MuiInputBase-root"]': {
+      height: 40,
+      fontSize: '0.875rem',
```
andreeimiron, line 32: Same here. Try to avoid 'rem' and use px instead

packages/fraction-model/src/main.jsx
```diff
@@ -0,0 +1,215 @@
...
+    }
+  };
+
+  /*
+   * Function to toggle correct answer
+   * @param {boolean} show contains boolean value to show correct answer
+   * */
+  toggleShowCorrect = (show) => {
+    this.setState({ showCorrect: show });
+    this.setState({ fractionModelKey: this.state.fractionModelKey + 1 });
```
andreeimiron, line 94: should be change to `this.setState({ showCorrect: show, fractionModelKey: this.state.fractionModelKey + 1 });`

andreeimiron (changes requested):

packages/fraction-model/configure/package.json
akshayhiwale: Updated

packages/fraction-model/controller/package.json
akshayhiwale: Updated

packages/fraction-model/configure/src/model-options.jsx
akshayhiwale: Updated

packages/fraction-model/configure/src/number-text-field.jsx
akshayhiwale: Updated

packages/fraction-model/src/main.jsx
akshayhiwale: Updated

andreeimiron (approved):

packages/fraction-model/configure/CHANGELOG.md
```diff
@@ -0,0 +1,742 @@
+# Change Log
```
andreeapescar, line 1: I think changelog should be empty, right?

packages/fraction-model/CHANGELOG.md
```diff
@@ -0,0 +1,784 @@
+# Change Log
```
andreeapescar, line 1: Changelog should be empty, right?

packages/fraction-model/configure/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "name": "@pie-element/fraction-model-configure",
+  "private": true,
+  "version": "1.0.0",
+  "description": "",
+  "main": "lib/index.js",
+  "module": "src/index.js",
+  "dependencies": {
+    "@material-ui/core": "^3.9.2",
+    "@pie-framework/pie-configure-events": "^1.3.0",
+    "@pie-lib/pie-toolbox": "1.29.1",
```
andreeapescar, line 11: Please use latest pie-lib/pie-toolbox

packages/fraction-model/controller/CHANGELOG.md
```diff
@@ -0,0 +1,647 @@
+# Change Log
```
andreeapescar, line 1: clean changelog please

packages/fraction-model/controller/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@pie-element/fraction-model-controller",
+  "private": true,
+  "version": "1.0.0",
+  "description": "",
+  "main": "lib/index.js",
+  "module": "src/index.js",
+  "dependencies": {
+    "@pie-lib/pie-toolbox": "1.29.1",
```
andreeapescar, line 9: update pie-lib/pie-toolbox please

packages/fraction-model/demo-fm.gif
andreeapescar, line 1: I really like the fact you created a gif for this, thank you! But I think that we don't want it in the element, it will just increase the size of the component.

packages/fraction-model/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": "pie-framework/pie-elements",
+  "version": "1.0.0",
+  "description": "",
+  "dependencies": {
+    "@material-ui/core": "^3.9.2",
+    "@pie-framework/pie-player-events": "^0.1.0",
+    "@pie-lib/pie-toolbox": "1.29.1",
```
andreeapescar, line 12: update pie-lib/pie-toolbox

andreeapescar (changes requested):

packages/fraction-model/demo-fm.gif
andreeimiron: Maybe we should just exclude this file from the build. Or let's keep it for now to see how it works.

andreeapescar: Ok let's keep it for now

packages/fraction-model/controller/CHANGELOG.md
akshayhiwale: Updated

packages/fraction-model/controller/package.json
akshayhiwale: Updated

packages/fraction-model/demo-fm.gif
akshayhiwale: Updated

andreeapescar (approved):

PR #2132, opened 2024-09-10
fix(inline-dropdown): add configs for toolbar inside inline-dropdown PD-4065

barta-adrian:
https://illuminate.atlassian.net/browse/PD-4065

### DIFF
--- a/packages/inline-dropdown/configure/src/defaults.js
+++ b/packages/inline-dropdown/configure/src/defaults.js
@@ -33,13 +33,13 @@ export default {
         image: { disabled: false },
       },
     },
-    toolbarInputConfiguration: {
+    responseAreaInputConfiguration: {
       audio: { disabled: true },
       video: { disabled: true },
       image: { disabled: true },
       table: { disabled: true },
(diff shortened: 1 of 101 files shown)


packages/inline-dropdown/configure/src/defaults.js
```diff
@@ -33,6 +33,14 @@ export default {
         image: { disabled: false },
       },
     },
+    toolbarInputConfiguration: {
```
andreeapescar, line 36: can we name this responseAreaInputConfiguration instead?

packages/inline-dropdown/configure/src/defaults.js
```diff
@@ -33,6 +33,14 @@ export default {
...
+      audio: { disabled: true },
+      video: { disabled: true },
+      image: { disabled: true },
+      table: { disabled: true },
+      'bulleted-list': { disabled: true },
```
andreeapescar, line 41: these need to be 'ol-list' and 'ul-list'

PR #2165, opened 2024-09-17
fix: update default values PD-3744

andreeimiron:
https://illuminate.atlassian.net/browse/PD-3744


packages/multi-trait-rubric/controller/src/defaults.js
```diff
@@ -1,16 +1,37 @@
...
     excludeZero: false,
+    halfScoring: false,
     maxPointsEnabled: true,
-    addScaleEnabled: true,
+    pointLabels: true,
+    scales: [],
+    standards: false,
+    visibleToStudent: true,
   },
   configuration: {
```
andreeapescar, line 13: I think there's a mistake here, so it might be just the right time to fix this.

defaults from controller should only have `model`, not `configuration`, because the controller doesn't have access in real like to `configuration`.
Also, it seems like it's not being used properly in packages/multi-trait-rubric/controller/src/index.js, let's fix this:

`export function createDefaultModel(model = {}) { return new Promise((resolve) => resolve({ ...defaults, ...model })); }`
`export const normalize = (question) => ({ ...defaults, ...question });`

packages/complex-rubric/controller/src/defaults.js
```diff
@@ -140,19 +176,28 @@ const rubriclessDefaultConfiguration = {
...
+      [RUBRIC_TYPES.SIMPLE_RUBRIC]: rubricDefaultModel,
+      [RUBRIC_TYPES.MULTI_TRAIT_RUBRIC]: multiTraitDefaultModel,
+      [RUBRIC_TYPES.RUBRICLESS]: rubriclessDefaultModel,
+    },
   },
   configuration: {
-    multiTraitRubric: multiTraitDefaultConfiguration,
-    simpleRubric: rubricDefaultConfiguration,
-    rubricless: rubriclessDefaultConfiguration,
+    rubricOptions: ['simpleRubric', 'multiTraitRubric', 'rubricless'],
```
andreeapescar, line 198: I think there's a mistake here, so it might be just the right time to fix this.

defaults from controller should only have `model`, not `configuration`, because the controller doesn't have access in real like to `configuration`.
In complex-rubric it looks like we're not using this, anyway.

andreeapescar (changes requested):

andreeapescar (approved):

PatriciaRomaniuc (approved):

PR #2168, opened 2024-09-20
Fix/pd 4029

arimieandreea:


packages/hotspot/src/hotspot/polygon.jsx
```diff
@@ -1,13 +1,16 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Line, Group } from 'react-konva';
+import { Line, Group, Rect } from 'react-konva';
```
PatriciaRomaniuc, line 3: I don't think you used this new imported component

PatriciaRomaniuc (approved):

PR #2194, opened 2024-10-02
fix(charting): disable edit labels in view mode PD-3928

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PD-3928

packages/charting/src/main.jsx
```diff
@@ -58,12 +58,19 @@ export class Main extends React.Component {
       studentNewCategoryDefaultLabel,
       showToggle,
       rationale,
-      correctedAnswer,
       correctAnswer,
       language,
+      env,
     } = model;
 
+    let { correctedAnswer } = model;
```
CarlaCostea, line 66: why are we not doing this in controller, mapping correctedAnswer there?

CarlaCostea (changes requested):

PR #2198, opened 2024-10-07
fix(fraction-model): change question property to prompt for TTS issue

akshayhiwale:
fix(fraction-model): change question property to prompt for TTS issue
https://illuminate.atlassian.net/browse/SC-32705

PR #2214, opened 2024-10-09
Feat/sc 32705

andreeapescar:
Akshay created this [PR in develop](https://github.com/pie-framework/pie-elements/pull/2198), but currently develop is "dirty" and I don't want to merge it in master.

So I'm opening a PR only with his changes directly in master.

"fix(fraction-model): change question property to prompt for TTS issue
https://illuminate.atlassian.net/browse/SC-32705"


packages/fraction-model/src/answer-fraction.jsx
```diff
@@ -19,10 +19,17 @@ export class AnswerFraction extends React.Component {
    * @returns updated answer change object
    */
   onValueChange = (key) => (event) => {
-    const { onAnswerChange, answers } = this.props;
+    let value = parseInt(event.target.value);
+    const { model, onAnswerChange, answers } = this.props;
     const newAnswers = { ...answers };
-
-    newAnswers[key] = parseInt(event.target.value);
+    let min = 1;
```
andreeimiron, line 25: min and max should be a const here

andreeimiron (approved):

packages/fraction-model/src/answer-fraction.jsx
```diff
@@ -19,10 +19,17 @@ export class AnswerFraction extends React.Component {
    * @returns updated answer change object
    */
   onValueChange = (key) => (event) => {
-    const { onAnswerChange, answers } = this.props;
+    let value = parseInt(event.target.value);
```
CarlaCostea, line 22: I would suggest adding a sanity check for event.target here to ensure it's defined and has a valid value

rakesh63760: Updated

packages/fraction-model/src/fraction-model-chart.js
```diff
@@ -221,13 +223,25 @@ const FractionModelChart = (props) => {
...
+    return <div ref={pieChartRef} className={classes.pieChartParentDiv}>{pieItems}</div>;
   };
 
   //Render bar or pie models as per model type
   if (modelType === 'bar') {
     return getBarFractionModel();
   } else if (modelType === 'pie') {
+    //Remove the last sector line of pie chart if parts per model is 1
+    useEffect(() => {
+      if (pieChartRef.current && partsPerModel === 1){
```
CarlaCostea, line 235: sanity check

packages/fraction-model/src/answer-fraction.jsx
```diff
@@ -19,10 +19,17 @@ export class AnswerFraction extends React.Component {
    * @returns updated answer change object
    */
   onValueChange = (key) => (event) => {
-    const { onAnswerChange, answers } = this.props;
+    let value = parseInt(event.target?.value);
```
CarlaCostea, line 22: please add also for the event

PR #2217, opened 2024-10-11
fix(categorize): add style to grid based on categpriesPerRow PD-4223

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/jira/software/c/projects/PD/issues/PD-4223

PR #2236, opened 2024-10-17
fix: preview prompt used for title and prompt to fix TTS issue

rakesh63760:
Preview prompt used for title and prompt to fix TTS issue

### DIFF
--- a/packages/fraction-model/src/main.jsx
+++ b/packages/fraction-model/src/main.jsx
@@ -5,6 +5,7 @@ import AnswerFraction from './answer-fraction';
 import { CorrectAnswerToggle } from '@pie-lib/pie-toolbox/correct-answer-toggle';
 import FractionModelChart from './fraction-model-chart';
 import { AlertDialog } from '@pie-lib/pie-toolbox/config-ui';
+import { PreviewPrompt } from '@pie-lib/pie-toolbox/render-ui';
 import cloneDeep from 'lodash/cloneDeep';
 
 export class Main extends React.Component {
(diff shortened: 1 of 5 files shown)


akshayhiwale (approved):

packages/fraction-model/src/main.jsx
```diff
@@ -121,15 +122,19 @@ export class Main extends React.Component {
...
     return (
       <div>
         <div className={classes.modelPreview}>
-          <h3 className={classes.titleContainer} dangerouslySetInnerHTML={{ __html: model.title }}></h3>
-          <p dangerouslySetInnerHTML={{ __html: model.prompt }}></p>
+          <h3 className={classes.titleContainer}>
+            <PreviewPrompt className="prompt" prompt={title} />
+          </h3>
+          <p>
+            <PreviewPrompt className="prompt" prompt={prompt} />
```
andreeimiron, line 136: you don't need to wrap `PreviewPrompt` in a `p` tag. You can add pass the tag you want for the text to the `PreviewPrompt` component (by default it's a `div`):
`<PreviewPrompt className="prompt" prompt={prompt} tagName="p" />`
same thing for the title

andreeimiron (changes requested):

packages/fraction-model/src/main.jsx
rakesh63760: made changes as per suggestion

andreeimiron (approved):

CarlaCostea:
https://illuminate.atlassian.net/browse/SC-32705

PR #2310, opened 2024-12-03
fix(image-cloze-association): preserve response ids when moved back to possibleResponses, Added fallback to generate new ids only when shifted responses lack an id PD-4538

CarlaCostea:
https://illuminate.atlassian.net/browse/PD-4538

### DIFF
--- a/packages/image-cloze-association/src/root.jsx
+++ b/packages/image-cloze-association/src/root.jsx
@@ -33,7 +33,7 @@ const styles = (theme) => ({
   },
 });
 
-class ImageClozeAssociationComponent extends React.Component {
+export class ImageClozeAssociationComponent extends React.Component {
   constructor(props) {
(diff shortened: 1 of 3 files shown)


packages/image-cloze-association/src/root.jsx
```diff
@@ -33,7 +33,7 @@ const styles = (theme) => ({
   },
 });
 
-class ImageClozeAssociationComponent extends React.Component {
+export class ImageClozeAssociationComponent extends React.Component {
```
PatriciaRomaniuc, line 36: Don't we already export the component at the bottom of the file?

CarlaCostea: for testing purposes, I needed direct access to the unwrapped class component —this change is only for enabling better unit testing.

PR #2312, opened 2024-12-03
feat(multiple-choice): add keydown support for selecting options PD-4411

barta-adrian:
https://illuminate.atlassian.net/browse/PD-4411


CarlaCostea (changes requested):
just wanted to bring up a small concern regarding the global keydown listener added 

there might be cases where this could interfere with other elements on the page? For example, if there are multiple elements like extended-text-entry or similar, pressing keys in one of them might accidentally trigger actions in the multiple-choice item, like selecting a checkbox. I could imagine scenarios where typing in a text field might unintentionally select a radio button or toggle something

PR #2321, opened 2024-12-09
fix(hotspot): apply hotspotColor when selectedColor not provided PD-4399

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/jira/software/c/projects/PD/issues/PD-4399

PR #2338, opened 2024-12-16
fix(math-inline): avoid rendering tooltip at start top, set flip fals…

arimieandreea:
…e PD-4529


arimieandreea:
https://illuminate.atlassian.net/jira/software/c/projects/PD/issues/PD-4529

PR #2350, opened 2024-12-19
fix(drag-in-the-blank): check alternates for feedback in evaluate PD-…

arimieandreea:
…3227


arimieandreea:
https://illuminate.atlassian.net/browse/PD-3227

PR #2354, opened 2024-12-20
fix(explicit-constructed-response): update regex to find response are…

arimieandreea:
…as mapped with 2 digits score PD-4535


arimieandreea:
https://illuminate.atlassian.net/browse/PD-4535

andreeimiron (approved):

PR #2373, opened 2025-01-14
fix(ECR): fix disable/enable spanish input for student view PD-4686

dianagociu:
https://illuminate.atlassian.net/browse/PD-4686

### DIFF
--- a/packages/explicit-constructed-response/configure/src/index.js
+++ b/packages/explicit-constructed-response/configure/src/index.js
@@ -118,23 +118,16 @@ export default class ExplicitConstructedResponse extends HTMLElement {
   }
 
   onConfigurationChanged(c) {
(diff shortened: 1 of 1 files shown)


CarlaCostea (commented):
@dianagociu, could this change potentially cause unintended updates? It seems the model might repeatedly update with the same configuration

dianagociu:
@CarlaCostea, The previous condition was never true because responseAreaInputConfiguration.inputConfiguration.characters.disabled gets modified as soon as the toggle is switched. Consequently, I removed the condition to ensure the model updates properly. Repeated updates with the same configuration won't occur because the toggles ensure the method is triggered only when switched, resulting in a different value each time.

andreeimiron (approved):

PR #2376, opened 2025-01-20
fix(image-cloze-association): chage icon to warning message and adapt…

arimieandreea:
… style PD-3872


arimieandreea:
https://illuminate.atlassian.net/browse/PD-3872

PR #2451, opened 2025-02-25
fix(explicit-constructed-response): get rid of local state as it shou…

arimieandreea:
…ld be updated based on session value PD-4695


arimieandreea:
https://illuminate.atlassian.net/browse/PD-4695

andreeimiron (approved):

PR #2580, opened 2025-05-13
fix(pie-models): set d3-array to resolution dep PD-5014

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PD-5014

PR #2877, opened 2026-01-20
fix(number-line): wrap content in g to attach the ref to DOM node PD-…

arimieandreea:
…5493


arimieandreea:
https://illuminate.atlassian.net/browse/PD-5493

PR #2904, opened 2026-02-11
fix(math-inline): get rid of unnessary container PD-5527

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PD-5527

PR #2907, opened 2026-02-13
fix(match): overwrite mui default colors to reflect correctness PD-5523

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PD-5523

PR #2912, opened 2026-02-18
fix(categorize, image-cloze-association): enable snap back visual onl…

arimieandreea:
…y for invalid drop zones PD-5495


arimieandreea:
https://illuminate.atlassian.net/browse/PD-5495

PR #2913, opened 2026-02-19
fix(extended-text-entry): send backgroundColor as props t editor PD-5548

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PD-5548

PR #2918, opened 2026-02-20
fix(multiple-choice): wait dom to be fully painted until calling rend…

arimieandreea:
…erMath requestAnimFrame PD-5619


arimieandreea:
https://illuminate.atlassian.net/browse/PD-5619

PR #2919, opened 2026-02-20
fix(drawing-response): avoid drawing when text element is moved PD-5593

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PD-5593

PR #2923, opened 2026-02-24
fix(hotspot): leave hover state while resizing PD-5601

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PD-5601

PR #2974, opened 2026-04-01
fix(math inline): avoid scroll position to be changed suring typing P…

arimieandreea:
…IE-15


arimieandreea:
https://illuminate.atlassian.net/browse/PIE-15

PR #2991, opened 2026-04-21
fix(math-templated): handle math keypad possition according to req, a…

arimieandreea:
…void kekpad move by adding content in input PIE-23


arimieandreea:
https://illuminate.atlassian.net/browse/PIE-23

PR #3009, opened 2026-05-06
fix: avoid keypad overlap the input when no below available space PIE…

arimieandreea:
…-113


arimieandreea:
https://illuminate.atlassian.net/browse/PIE-23

PR #3012, opened 2026-05-07
fix(hotspot): avoid setting hovercolor in view and eval mode PIE-435

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PIE-435

PR #3016, opened 2026-05-15
fix(math-templated): remove response are one by at backslash PIE-550

arimieandreea:


arimieandreea:
https://illuminate.atlassian.net/browse/PIE-550 (deletion part)

PR #3024, opened 2026-05-25
fix(inline-dropdown): run validation when a response area toolbar com…

arimieandreea:
…ponent is closed PIE-574


arimieandreea:
https://illuminate.atlassian.net/browse/PIE-574

PR #3042, opened 2026-06-16
fix: nulling root in case consumer triggers recreation of the compone…

arimieandreea:
…nt on update PIE-672


arimieandreea:
https://illuminate.atlassian.net/browse/PIE-672

PR #3051, opened 2026-06-18
fix: handle webcomponent lifecycle on every element that misses it PI…

arimieandreea:
…E-703


arimieandreea:
https://illuminate.atlassian.net/browse/PIE-703

PR #3071, opened 2026-07-06
fix(categorize): compute allowmultiple based on response when missing…

arimieandreea:
…, get rid of defaults value PIE, add tests PIE-723


arimieandreea:
https://illuminate.atlassian.net/browse/PIE-723

PR #3080, opened 2026-07-28
fix(multiple-choice): maxSelections should not affect single choices,…

arimieandreea:
… repair tests in src PIE-432


arimieandreea:
Within this pr we avoid taking into consideration maxSelections for single choices (radio buttons), we needed to add this verification in validate such as in authoring cannot add more correct responses than maxSelections when the case and repair and rename all the tests according to jest regex because tests from src were never executed

arimieandreea:
https://illuminate.atlassian.net/browse/PIE-432

PatriciaRomaniuc (approved):

